PR #2, opened 2022-06-28
Add pre-commit hooks with linters

nhuet:
- Use pre-commit to check that files respect some format (black, isort, ...)
- Add a step checking these hooks in github workflow
- Format all files to respect the hooks

You can now use 
- pre-commit on all files once in a while with `pre-commit run -a` 
- pre-commit at each commit with `pre-commit install`


nhuet:
Let us wait for PR #1 before merging this one.

PR #6, opened 2022-07-04
Fix issues when gurobi not available

nhuet:
Some tests were not working if launched with gurobi not installed. 
We fix the issues and add a job launching the test suite without gurobi (for only one entry of matrix os/python-version).


nhuet:
All tests are OK, except for [an entry of the strategic matrix](https://github.com/airbus/discrete-optimization/runs/7181512097?check_suite_focus=true). It seems to be a random error as though the randomness was not fixed although we tried to do it already, by fixing the random seed with both modules `random` and `numpy`.

g-poveda:
Ok No problem, maybe one day we'll use ```from __future__ import annotation``` instead of typing the object with double quote ?

PR #8, opened 2022-07-04
Restore @njit decorator for sgs_fast_preemptive_minduration()

nhuet:
A @njit decorator was commented on a "fast" function.
Probably a debugging manipulation that have been commited by mistake.


nhuet:
Same comment as in [PR #6](https://github.com/airbus/discrete-optimization/pull/6#issuecomment-1173876764).
A job failed probably because fixing the randomness is not perfectly done.

PR #11, opened 2022-07-05
Allow more time for rcpsp_ms to rcpsp solver

nhuet:
From time to time, the tests in tests/rcpsp_multiskill/test_rcpsp_ms_to_rcpsp.py fail beacause they do not find a solution in time. We try to relax a bit the time conditions on the solver, going partially back on changes from commit df20a0e.

### DIFF
--- a/discrete_optimization/rcpsp_multiskill/multiskill_to_rcpsp.py
+++ b/discrete_optimization/rcpsp_multiskill/multiskill_to_rcpsp.py
@@ -42,7 +42,7 @@ class MultiSkillToRCPSP:
         one_worker_type_per_task: bool = False,
     ):
         params_cp = ParametersCP(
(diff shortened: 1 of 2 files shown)


discrete_optimization/rcpsp_multiskill/multiskill_to_rcpsp.py
```diff
@@ -42,7 +42,7 @@ def construct_rcpsp_by_worker_type(
         one_worker_type_per_task: bool = False,
     ):
         params_cp = ParametersCP(
-            time_limit=0.5,
+            time_limit=30,
```
g-poveda, line 45: This value Can be a bit conservative, however it seems the computation time of test unit don't explode so Ok !

PR #19, opened 2022-07-22
[codereview.doctor] Remove "if" conditions that always evaluate to True or False

nhuet:
These issues have been automatically detected by codereview.doctor

- Either these trivial conditions are the remaining of development process  => this is fine to remove them
- or they should not be evaluated always to `True `(respectively to `False`) => we should correct them e.g. by removing the `and False` or `or True` that are in the conditions

This PR is removing simply the conditions so that code behaviour is left unchanged but we need a review by someone that actually know what is behind the code (@POPOGO ?)

### DIFF
--- a/discrete_optimization/pickup_vrp/solver/lp_solver_pymip.py
+++ b/discrete_optimization/pickup_vrp/solver/lp_solver_pymip.py
@@ -1646,8 +1646,6 @@ def rebuild_routine(
                         backup_min_dist = cost
             if len(edge_out_of_interest) == 0:
                 return None
-            if min_out_edge is None and False:
-                return None
             if min_component is None:
                 return None
             len_this_component = len(paths_component[min_component])
--- a/discrete_optimization/pickup_vrp/solver/lp_solver_pymip.py
+++ b/discrete_optimization/pickup_vrp/solver/lp_solver_pymip.py
@@ -1968,6 +1968,7 @@ def update_model_lazy(
     components_global,
     edges_in_all_vehicles,
     edges_out_all_vehicles,
+    do_constraint_on_order=False,
 ):
     len_component_global = len(components_global)
     list_constraints = []
@@ -2018,7 +2019,7 @@ def update_model_lazy(
                         >= 1
                     )
                 ]
-    if len_component_global > 1 and False:
+    if len_component_global > 1 and do_constraint_on_order:
         constraints_order = {}
         try:
             variable_order = lp_solver.variable_order
--- a/discrete_optimization/vrp/solver/lp_vrp_iterative_pymip.py
+++ b/discrete_optimization/vrp/solver/lp_vrp_iterative_pymip.py
@@ -905,19 +905,18 @@ def update_model_2(
     if len_component_global > 1:
         print("Nb component : ", len_component_global)
         for s in components_global:
-            if True:
-                edge_in_of_interest = [
-                    e
-                    for n in s[0]
-                    for e in edges_in_customers[n]
-                    if e[0][1] not in s[0] and e[1][1] in s[0]
-                ]
-                edge_out_of_interest = [
-                    e
-                    for n in s[0]
-                    for e in edges_out_customers[n]
-                    if e[1][1] not in s[0] and e[0][1] in s[0]
-                ]
-                model.add_constr(xsum([x_var[e] for e in edge_in_of_interest]) >= 1)
-                model.add_constr(xsum([x_var[e] for e in edge_out_of_interest]) >= 1)
+            edge_in_of_interest = [
+                e
+                for n in s[0]
+                for e in edges_in_customers[n]
+                if e[0][1] not in s[0] and e[1][1] in s[0]
+            ]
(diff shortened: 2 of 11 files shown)


discrete_optimization/pickup_vrp/solver/lp_solver_pymip.py
```diff
@@ -2010,45 +2008,6 @@ def update_model_lazy(
                         >= 1
                     )
                 ]
-    if len_component_global > 1 and False:
```
g-poveda, line 2013: Here : the piece of code should still exist. the False has been added to improve performance for some experiments and I should have removed it. We should pass an argument to the function, by default False and use it in this condition.

discrete_optimization/vrp/solver/lp_vrp_iterative_pymip.py
```diff
@@ -406,82 +406,81 @@ def solve(self, **kwargs):
                 best_solution_objective_rebuilt = obj
                 best_solution_rebuilt_index = iteration
             iteration += 1
-            if len(component_global_all[0]) > 1 or True:
```
g-poveda, line 409: same comment as the lp_solver_pymip (which is very similar). for this one, better to introduce an argument to the function

g-poveda (commented):
Approved except for 2 piece of code to change

discrete_optimization/pickup_vrp/solver/lp_solver_pymip.py
nhuet: So by default, this code block must be ignored? (with the argument at False)
Or the opposite (since you say that you should have removed the `and False`) ?
What could be the name of the variable holding this value ? `thorough_update` ? So that if True, we pass potentially in this block and if False not?

discrete_optimization/vrp/solver/lp_vrp_iterative_pymip.py
nhuet: Same question: by default do we want to bypass the block code?

discrete_optimization/pickup_vrp/solver/lp_solver_pymip.py
nhuet: I tried something (see the 2 last commits). I put by default True because you said that you should have removed the "and False". But you said also that the argument should be False by default so i am not so sure about it...

discrete_optimization/vrp/solver/lp_vrp_iterative_pymip.py
g-poveda: Ok my comment was a bit wrong, re-reading the full code, the correct fix is to keep the "if len(component_global_all[0]) > 1" condition, and remove the True

discrete_optimization/pickup_vrp/solver/lp_solver_pymip.py
g-poveda: I don't see when "thorough_update" is used in the core of the function ? 
`len_component_global = len(components_global)
    list_constraints = []
    if len_component_global > 1:
        print("Nb component : ", len_component_global)
        for s in components_global:
            # print(s)
            edge_in_of_interest = [e for n in s[0]
                                   for e in edges_in_all_vehicles[n]
                                   if e[1][0] not in s[0]
                                   and e[1][1] in s[0]]
            edge_out_of_interest = [e for n in s[0]
                                    for e in edges_out_all_vehicles[n]
                                    if e[1][0] in s[0]
                                    and e[1][1] not in s[0]]
            if not any(problem.target_vehicle[v] in s[0] for v in problem.origin_vehicle):
                list_constraints += [lp_solver.model.add_lazy_constr(mip.quicksum([lp_solver.variable_decisions["variables_edges"][e[0]][e[1]]
                                                                                   for e in edge_out_of_interest]) >= 1)]
            if not any(problem.origin_vehicle[v] in s[0] for v in problem.origin_vehicle):
                list_constraints += [lp_solver.model.add_lazy_constr(mip.quicksum([lp_solver.variable_decisions["variables_edges"][e[0]][e[1]]
                                                                                   for e in edge_in_of_interest]) >= 1)]
    if len_component_global > 1 and thorough_update:
        constraints_order = {}
        try:
            variable_order = lp_solver.variable_order
`

g-poveda: and by default, thorough_update wouls be false, so the behaviour would be the same as the original code. We can rename this parameter with "do_constraint_on_order", i'll have to documentate it later.

discrete_optimization/vrp/solver/lp_vrp_iterative_pymip.py
nhuet: Taken into account

discrete_optimization/pickup_vrp/solver/lp_solver_pymip.py
nhuet: Done

PR #22, opened 2022-07-22
[coderview.doctor] Use dataclass instead of NamedTuple

nhuet:
These improvements have been automatically detected by coderview.doctor.

See https://codereview.doctor/features/python/best-practice/avoid-named-tuple-use-dataclass
for more explanation.


g-poveda (commented):
Ok ! coherent with the PR !

PR #24, opened 2022-07-22
[coderview.doctor] Use Optional instead of Union

nhuet:
These improvements have been automatically detected by coderview.doctor.


g-poveda (approved):
Limpide

PR #26, opened 2022-08-30
Remove unnecessary prints

g-poveda:
Simply delete some print, the remaining one will have to be replaced by loggers.


nhuet:
Ok for me.

PR #27, opened 2022-08-30
Avoid duplicates by calling look_for_solver_class in look_for_solver

nhuet:


g-poveda (approved):
Ok conform to commit description, good refactoring

PR #29, opened 2022-09-02
Remove solve_lns() methods from CP and LP solvers

nhuet:
We can instead use the generic lns solvers `discrete_optimization.generic_tools.lns_cp.LNS_CP` or `discrete_optimization.generic_tools.lns_mip.LNS_MILP`.


g-poveda:
No pb, remove deprecated solvers as expected

PR #32, opened 2022-09-05
Examples Fix and refactoring

g-poveda:
- Removed the no longer existing lns solver test in gphh_facility_example
- fix a minor bug found in routing ortoolsolver 
- renamed some examples scripts (test->run)

### DIFF
--- a/discrete_optimization/facility/solvers/gphh_facility.py
+++ b/discrete_optimization/facility/solvers/gphh_facility.py
@@ -197,6 +197,7 @@ class GPHH(SolverDO):
     ):
         self.training_domains = training_domains
         self.domain_model = domain_model
+        self.params_gphh = params_gphh
         if params_gphh is None:
             self.params_gphh = ParametersGPHH.default()
         self.set_feature = self.params_gphh.set_feature
--- a/discrete_optimization/facility/solvers/gphh_facility.py
+++ b/discrete_optimization/facility/solvers/gphh_facility.py
@@ -197,9 +197,10 @@ class GPHH(SolverDO):
     ):
         self.training_domains = training_domains
         self.domain_model = domain_model
-        self.params_gphh = params_gphh
         if params_gphh is None:
             self.params_gphh = ParametersGPHH.default()
+        else:
+            self.params_gphh = params_gphh
         self.set_feature = self.params_gphh.set_feature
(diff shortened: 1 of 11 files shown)


discrete_optimization/facility/solvers/gphh_facility.py
```diff
@@ -197,6 +197,7 @@ def __init__(
     ):
         self.training_domains = training_domains
         self.domain_model = domain_model
+        self.params_gphh = params_gphh
         if params_gphh is None:
             self.params_gphh = ParametersGPHH.default()
```
nhuet, line 202: I would rather do:
```python
        if params_gphh is None:
            self.params_gphh = ParametersGPHH.default()
        else:
            self.params_gphh = params_gphh
```

This is more explicit imo

discrete_optimization/pickup_vrp/solver/ortools_solver.py
```diff
@@ -258,10 +258,13 @@ def distance_callback(from_index, to_index):
         if include_equilibrate_charge:
             charge = routing.RegisterTransitCallback(lambda i, j: 1)
             charge_constraint = kwargs.get("charge_constraint", {})
+            max_charge = len(self.problem.list_nodes)
+            if all(charge_constraint[i][1] is not None for i in charge_constraint):
+                max_charge = max([charge_constraint[i][1] for i in charge_constraint])
```
nhuet, line 263: Same here:
```python
            if all(charge_constraint[i][1] is not None for i in charge_constraint):
                max_charge = max([charge_constraint[i][1] for i in charge_constraint])
            else:
                max_charge = len(self.problem.list_nodes)
```

examples/knapsack/multiscenario_knap.py
```diff
@@ -60,10 +61,11 @@ def initialize_multiscenario():
         temperature_handler=TemperatureSchedulingFactor(1000, res, 0.99),
         mode_mutation=ModeMutation.MUTATE,
     )
-    sa.solve(
-        initial_variable=solution,
-        nb_iteration_max=1000000,
-        pickle_result=False,
+    import logging
+
+    logging.basicConfig(level=logging.DEBUG)
```
nhuet, line 66: Import at the beginning of the file.

PR #33, opened 2022-09-06
Refactor milp solvers

nhuet:
We gather `solve()` and some utility methods in 2 subclasses of `MilpSolver`: `GurobipyMilpSolver` and `PymipMilpSolver` depending if the solver wraps a `gurobipy.Model` or a `mip.Model`.

The idea is to have one version of `solve()` by type of solver which has all parameters correctly passed to `self.model`. Formerly, the solve was not always using all parameters

We can also put `retrieve_solutions` in common inside each module by using the utility functions. We improve it by always looping on solutions if `parameters_milp.retrieve_all_solution` is `True`.


g-poveda:
Good and clear improvement of the API.

PR #35, opened 2022-09-08
Refactor CP solvers

nhuet:
They actually all wrap a minizinc instance, so we create a corresponding generic class that implements the solve method they can share. This simplifies the maintenance and ensure that all parameters from parameters_cp are correctly used everywhere.

As some solvers need to raise a warning instead of an error when self.instance.solve() crashes, we add a silent_solve_error attribute which is False by default.

Vrp CP solver was not mature so we decided to remove it.


discrete_optimization/coloring/solvers/coloring_cp_solvers.py
```diff
@@ -161,53 +161,33 @@ def retrieve_solutions(self, result, parameters_cp: ParametersCP) -> ResultStora
...
+                objectives.append(result.objective)
         for k in range(len(colors)):
             sol = [
                 colors[k][self.index_nodes_name[self.nodes_name[i]]] - 1
                 for i in range(self.number_of_nodes)
             ]
             color_sol = ColoringSolution(self.coloring_problem, sol)
             fit = self.aggreg_sol(color_sol)
-            solutions_fit += [(color_sol, fit)]
+            solutions_fit.append((color_sol, fit))
```
g-poveda, line 183: why ?! but ok

g-poveda (approved):
No pb, shared solve function for the minizinc based cp solver proves to be a working idea.

PR #41, opened 2022-09-15
Update readme

nhuet:
- install process from pip (to be ready for the release)
- update existing notebooks (or that will exist in the near future)
- remove reference to ttp (not included in the current version)
- adapt readme so that it can be used as a long description in PyPI and be still understandable ("directory" -> "directory of the repository")


g-poveda:
Ok no pb

PR #43, opened 2022-09-15
Add copyright for source files

### LINKED ISSUE #14: Add copyright in the source code
-SInce it is a subproject of scikit-decide we can use the same template.
-For minizinc models already having copyright, add that it has been adapted by airbus

nhuet:
We add copyright to all python files, mimicking what is done for [scikit-decide](https://github.com/airbus/scikit-decide).
We also add copyright to minizinc files, mentioning the originals they have been adapted from. 

Fix #14.


g-poveda:
Ok thanks ! just remains the few files we discussed.

PR #48, opened 2022-09-29
Check minizinc binary version and bump minizinc-python version

nhuet:
- require version 2.6 for minizinc binary
- check at import of discrete_optimization package that
  - minizinc is installed and reachable from PATH
  - its version is greater or equal to 2.6.0
- bump minizinc-python to be able to do this check easily from python


galleon (approved):
LGTM

PR #50, opened 2022-10-10
Fix "Fetch data for tests" workflow step for windows

nhuet:
Since PR #48, minizinc is imported in
`discrete_optimization.__init__.py` to check the binary version, so even when fetching tests data.

On windows we need to redefine the `PATH` properly so that minizinc binary is found (while on mac and linux, having it defined at "Add MiniZinc to PATH" step seems sufficient).

The failure happens only if datasets cache retrieval fails itself. See this [job](https://github.com/nhuet/discrete-optimization/actions/runs/3154433525/jobs/5132021850) for instance.


g-poveda:
Ok fix a problem that appeared lately :)

PR #52, opened 2022-10-10
Update actions to use node 16

nhuet:
Actions running on node 12 will not work anymore in summer 2023 and are
now raising warnings:
https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/

So:
- we update when possible actions version to ones using node 16 instead of 12.
- we stop using deprecated pre-commit/action and replace it by simple call to pre-commit in a shell (mimicking pre-commit/action@v3)
- we replace softprops/action-gh-release by ncipollo/release-action@v1 which is using node 16

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -22,7 +22,15 @@ jobs:
       - uses: actions/setup-python@v4
         with:
           python-version: "3.8"
-      - uses: pre-commit/action@v3.0.0
+      - name: install pre-commit
+        run: python -m pip install pre-commit
+      - name: get cached pre-commit hooks
+        uses: actions/cache@v3
+        with:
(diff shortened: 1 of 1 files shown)


nhuet:
@g-poveda : Now that the workflow is working again, can you relaunch it for this PR? Thanks.

g-poveda:
@nhuet i don't find anymore where to launch the workflow.. https://github.com/airbus/discrete-optimization/tree/nht/wf_node_16 gives me a error 404

nhuet:
Actually amending the last commit (with nothing but a new commit date) and force pushing made the trick :) ! So nevermind relaunching it.

g-poveda:
i was not able to find the relaunch button pfff. but everything is green,

PR #56, opened 2022-10-17
Build and deploy the documentation

nhuet:
We use sphinx to build the doc and deploy it with github pages.
- documentation stuff is in docs/ directory
- source files are in docs/source
- explanations about how to set and build the doc are in docs/source/contribute.md
- existing notebooks are listed automatically in notebooks page of the doc
- in github workflow, 
    - we try to build the doc at each push => one can see it by downloading the corresponding artifact, 
      unzip it and run `python -m http.server` in the extracted directory.
    - if pushed on default branch, we deploy it (in {branch_name}/ surbdirectory of gh-pages branch)
    - when tagging a release (vx.y.z) we deploy also it (in x.y.z subdirectory of gh-pages branch)
    - after each doc deploy, a versions.json file at root of gh-pages branch is updated with the list of available doc versions in order to enable a menu allowing to switch between versions
    - the doc is then available at https://{username}.github.io/discrete-optimization, so after the PR at https://airbus.github.io/discrete-optimization
- You can see an example of the rendered doc with this workflow here: https://nhuet.github.io/discrete-optimization (fake releases have been done to see how to switch between versions of the doc)


.github/workflows/build.yml
```diff
@@ -206,7 +206,136 @@ jobs:
...
+        run: |
+          ${{ env.minizinc_install_cmdline }}
+      - name: Add MiniZinc to PATH
+        run: |
+          echo ${{ env.minizinc_path }}
+          echo ${{ env.minizinc_path }} >> $GITHUB_PATH
+      - name: Get pip cache dir
+        id: get-pip-cache-dir
+        run: |
+          echo "::set-output name=dir::$(pip cache dir)"
```
galleon, line 282: This is deprecated

galleon (dismissed):
One `set-out` still remaining

.github/workflows/build.yml
nhuet: Yep. Wanted to fix it in #57 . But i could also do it there.

galleon: LGTM

galleon: LGTM

galleon: LGTM

nhuet: @galleon it is still stating "changes requested", do you know why?

PR #57, opened 2022-10-17
Use new way of setting output parameter

nhuet:
The command set-output is now deprecated and we have now to use environment files to define step outputs:
https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/


galleon:
LGTM

PR #58, opened 2022-10-17
Use pyproject.toml instead of setup.py

nhuet:
This configuration file is now (since 2016) pushed to be the standard for packaging Python project, as stated here:

- packaging guide: https://packaging.python.org/en/latest/flow/#the-configuration-file
- PEP defining it: https://peps.python.org/pep-0518/
- setuptools (one of the tool that can be used for packaging) documentation: https://setuptools.pypa.io/en/latest/userguide/quickstart.html#basic-use


g-poveda:
LGTM since more standard and well tested as discussed privately

PR #61, opened 2022-10-18
Use new feature caching pip cache with setup-python action

nhuet:
It simplifies a bit the workflow as we remove 2 steps in general (getting the pip cache path which depends on plateform and then using the cache action itself).

However the action relies on the existence of "requirements.txt" files or at least of a file which manage the required dependencies.
Thus for "linter" and "build" job, we have a small step creating this file (respectively with only pre-commit, or build as defined dependencies).


g-poveda:
lgtm

PR #63, opened 2022-10-20
New knapsack solver based on cpmpy library

g-poveda:
- Adding a wrapper solver for knapsack using the CP modeling library in python https://github.com/CPMpy/cpmpy
- Adding example to run and unit test
- adding cpmpy as dependency

### DIFF
--- a/examples/knapsack/knapsack_cpmyp.py
+++ b/examples/knapsack/knapsack_cpmyp.py
@@ -1,0 +1,42 @@
+import logging
+
+from discrete_optimization.generic_tools.cp_tools import ParametersCP
+from discrete_optimization.generic_tools.lp_tools import MilpSolverName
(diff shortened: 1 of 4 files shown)


examples/knapsack/knapsack_cpmyp.py
```diff
@@ -1,3 +1,6 @@
+#  Copyright (c) 2022 AIRBUS and its affiliates.
+#  This source code is licensed under the MIT license found in the
+#  LICENSE file in the root directory of this source tree.
 import logging
 
 from discrete_optimization.generic_tools.cp_tools import ParametersCP
```
nhuet, line 6: Not very important, but should be part of previous commit.

nhuet (approved):
Ok for me

PR #66, opened 2022-10-21
Use pyproject.toml file for mypy configuration

nhuet:
Now that we have a pyroject.toml, we should use it for tools option, here mypy.

As "*" character is not permitted in toml keys, and since the list of dependencies without typing support is not yet stable, we ignore missing import instead for every packages.


g-poveda:
lgtm

PR #67, opened 2022-10-21
Fix a bug when calling get_data_available without data_home existing

nhuet:


g-poveda (commented):
what is the most correct way here, do try:except or use os.path.exists() ?

g-poveda:
LGTM

PR #68, opened 2022-10-21
Fix dependencies of the library and test importing each module correctly

nhuet:
We were missing scipy dependency needed in a lot of modules (espceially rcpsp ones).
Indeed test optional dependencies were installing it. So

- we fix pyproject.toml
- we add a step in workflow which is checking the imports *before* installing the optional dependencies


g-poveda:
LGTM

PR #72, opened 2022-11-04
Fix github+binder links for notebooks when tagging a release

nhuet:
The environment variables needed to generate the proper links were not defined when tagging a release. 
(In that case, $GITHUB_REF starts with "refs/tags")


g-poveda:
Ok to be checked by @nhuet it works as expected.

PR #73, opened 2022-11-04
Update automatically binder environment when pushing on default branch or tagging a release

nhuet:
The philosophy of binderhub is that "building the same commit at multiple points in time will result in the same image." (cf https://github.com/jupyterhub/binderhub/issues/683). 
That means that if we specify (in environment.yml  from binder branch) the dependency in discrete-optimization only to be the master version of the repository, the environment will not be rebuilt when master changes, as the environment.yml (and thus the branch binder) will stay the same (ie with an old version of the library). 
So the solution chosen is to modify the environment.yml to point not anymore on master branch but directly on the last commit of the master branch (by specifying the sha1)

So the resulting workflow is:

- When pushing on default branch:
  - we get the new sha1 of the last commit
  - we update environment.yml on binder branch with this sha1
  - we trigger a build of the new environment (to avoid waiting time to future users of notebooks)
  - we point to this branch binder in binder links for notebooks in doc (as before)
- When tagging a release:
  - we get the version number to be released
  - we update the environment.yml so that discrete-optimization is to be downloaded from pypi (or test.pypi for forks) with the proper version
  - we commit this new environment on binder branch and get the corresponding sha1 --> this is the reference for the release binder environment
  - we revert and push so that
    - the binder branch still points to an environemnt with library code from default branch (the development version)
    - the release environment is in the history of the remote repository and will be accessible from binderhub
  - we point to the stored reference for the release binder environment (the sha1 of the intermediate commit on binder branch) in binder links for notebooks in doc
  - we trigger a build of the environment for binder branch (as the branche advanced by 2 commits even though with same specs as before)
  - we trigger a build of the release environment (but only once we finished deploying on PyPI or TestPyPI)

Thus, the history of the binder branch will be something like  "dev version" > "release n version" > "dev version" > "dev version (new push on master)" >... > "release n+1 version" > ...


g-poveda:
LGTM. Ok, following your comment you can test it now that #72  is merged.

nhuet:
I rebased and pushed another tag v0.1.9. It worked as expected:
-  The generated doc is available [here](https://nhuet.github.io/discrete-optimization/0.1.19/notebooks.html) with the link for the knapsack notebook working properly.
- The opening in binder is quite fast because as you will see in logs the image was already built (thanks to the github workflow)
- The notebook is running without errors
-  If typing in the notebook:
  ```  
  import discrete_optimization
  discrete_optimization.__version__
  ```
  we can see that we install the proper version in the environment (here version 0.1.9 from test.pypi).

g-poveda:
lgtm

PR #74, opened 2022-11-07
Modify prerequisites text for binder users

nhuet:
The prerequisites are useful only for local users as everything has already been done for binder environment.


g-poveda (commented):
Ok !

PR #75, opened 2022-11-07
Simplify readme by pointing to online doc

nhuet:
Avoid duplicate instructions and will ease future maintenance of the installation doc. (Readme will probably not change anymore.)

This is much like what has been done for scikit-decide.


g-poveda:
lgtm

PR #79, opened 2022-11-18
Update minizinc install command line for ubuntu

nhuet:
Works even when FUSE is not available.


g-poveda:
Ok, now working on ubuntu, lgtm

PR #82, opened 2022-11-18
Correct minizinc install command lines in doc for Linux without FUSE

nhuet:
We add the lib/ directory to `LD_LIBRARY_PATH` to avoid issues with gecode.


g-poveda:
Thanks !!

PR #89, opened 2022-12-02
New attribute in CPSolvers : status_solver

g-poveda:
We add the possibility to store status of the solver, such as (sat, unsat, unknown, optimal). Interesting to retrieve for computational experiments.


g-poveda:
very small change, no bug expected

PR #90, opened 2022-12-05
MSLIB (+ polishing MSPSP) integration

g-poveda:
New dataset is now integrated in the lib, 
-It is the MSLIB from [Snauwaert, J., & Vanhoucke] available in https://www.projectmanagement.ugent.be/research/project_scheduling/MSRCPSP.
-Fetching function + parsing to discr-opt library and one example of solve is provided.

- Some fixes on the previous dataset included (MSPSP) are also included because it was conflicting with MSLIB dataset.


g-poveda:
Ok

PR #94, opened 2022-12-16
Robustify create-binder-env job

nhuet:
When relaunching a workflow where the action had previously run successfully, the "git commit" will have an non-null exit-code as there is nothing new to commit. So we check that a diff exists before doing a git commit.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -312,7 +312,12 @@ jobs:
             # Commit new environment
             git config user.name "Actions"
             git config user.email "actions@github.com"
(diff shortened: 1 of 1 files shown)


nhuet:
I tested that it actually prevents failing after a relaunch of whole workflow on default branch. 
This is done [here](https://github.com/nhuet/discrete-optimization/actions/runs/3712338529). 
You can see that on the second attempt, the create-binder-env job is still green :heavy_check_mark:

g-poveda:
Thanks to solve this issue !

PR #95, opened 2023-01-19
Fix mypy issues in knapsack

nhuet:
We solve issues in knapsack found by 

```
mypy discrete_optimization/knapsack
```

(essentially lack of typing or mistyping, or None not properly handled)

and other issues more generic (impacting other packages) highlighted by mypy errors in knapsack, essentially lack of return types.


g-poveda:
LGTM, typing completed for all knapsack and other side modules.

PR #97, opened 2023-01-19
Make mypy stricter

nhuet:
We use options
- enforcing typing all functions/methods/classes
- parsing also functions/methods/classes only partially typed (up to now, they were discarded by mypy so that issues were not seen)


g-poveda:
RAS

PR #102, opened 2023-01-19
Fix mypy issues in tsp

nhuet:


g-poveda:
The bugs in unit test are quite self explanatory, a bug was introduced about the initialisation permutation attribute

g-poveda:
Ok bug solved

PR #103, opened 2023-01-19
Fix mypy issues in vrp

nhuet:


discrete_optimization/vrp/mutation/mutation_vrp.py
```diff
@@ -175,7 +180,15 @@ def __init__(
...
         self.index_to = index_to
 
-    def apply_local_move(self, solution: VrpSolution) -> Solution:
+    def apply_local_move(self, solution: VrpSolution) -> VrpSolution:  # type: ignore # avoid isinstance checks for efficiency
+        if (
+            solution.length is None
+            or solution.lengths is None
+            or solution.capacities is None
+        ):
+            raise RuntimeError(
```
g-poveda, line 189: Warning, this test can lead to some inefficiencies in computation time.

PR #108, opened 2023-02-06
Update pickup_vrp API to be in par with the rest of the library

nhuet:
- We complete GPDPSolution class
- We convert output of pickup_vrp solvers so that `solve()` does return a `ResultStorage`.
- We fix various issues found by mypy
- Add tests for this new api

What this PR does not solve:
- LP solvers `LinearFlowSolverVehicleType` are unfinished
- in case of usecase with clusters, the output is perhaps not accurate (but this is not related to the change of API)


g-poveda:
as discussed, this PR is satisfactory and we identified some things to code further (better satisfy function for gpdp problem)

PR #109, opened 2023-02-14
ASP Solver for coloring problem binded to the library

g-poveda:
Using clingo library, we solve coloring problem using answer set programming paradigm.
It gives very good performance in the coloring benchmark examples.
A new plotting function for coloring solution has also been implemented and used in several existing scripts.

### DIFF
--- a/tests/coloring/test_coloring.py
+++ b/tests/coloring/test_coloring.py
@@ -1,7 +1,6 @@
 #  Copyright (c) 2022 AIRBUS and its affiliates.
 #  This source code is licensed under the MIT license found in the
 #  LICENSE file in the root directory of this source tree.
-
 import random
 
 import numpy as np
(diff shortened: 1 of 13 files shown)


tests/coloring/test_coloring.py
```diff
@@ -65,7 +64,7 @@ def test_solvers():
     assert coloring_model.graph.nodes_name is not None
     solvers = solvers_map.keys()
     for s in solvers:
-        if s == ColoringLP and not gurobi_available:
+        if s == ColoringLP:
```
nhuet, line 67: Why do we skip this solver evevn when shipping gurobi now? It was supposed to work up to now.
(And it will decrease the coverage i suppose)

nhuet (commented):
It would be nice to add some tests for this new ASP solver in tests directory :p

(And I am not a big fan of the way the commit messages are displayed because of too long first line or so. But not a big deal for now)

g-poveda:
Hi @nhuet i added a test, though it was already tested in tests/coloring/test_coloring.py/test_solvers(), because ASP solver is listed in the coloring_solvers module

PR #110, opened 2023-02-15
Add new generic CplexMilpSolver for linear programming using docplex library

g-poveda:
- added also a RCPSP CPLEX solver and associated examples
To be able to use it, one needs a have a commercial licence of CPLEX, therefore we don't include it in unit test.

In a long term horizon, we could think about coding "add_var", "add_constr".. function for all the linear programming library. So that for a given problem we would have a unique implementation for the LP model.

### DIFF
--- a/discrete_optimization/generic_tools/lp_tools.py
+++ b/discrete_optimization/generic_tools/lp_tools.py
@@ -5,7 +5,7 @@
 import logging
 from abc import abstractmethod
 from enum import Enum
-from typing import Any, Optional
+from typing import Any, List, Optional
 
 import mip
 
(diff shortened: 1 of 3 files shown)


nhuet (commented):
Once again, a new test would be appreciated for this new solver ;)

nhuet:
> In a long term horizon, we could think about coding "add_var", "add_constr".. function for all the linear programming library. So that for a given problem we would have a unique implementation for the LP model.

Indeed, like `get_obj_value_for_ith_solution` and `get_var_value_for_ith_solution` which are there to retrieve the variables and objectives. Good idea!

nhuet:
Maybe you shoud open an issue about that just to not forget it later.

PR #113, opened 2023-02-17
Replace addConstr() by newer addLConstr() available for gurobi >= 9.0

nhuet:
- this remove some warnings that pops up in pycharm
- this is supposed to be faster
- this has been introduced in gurobi 9
- we add the minimal version for gurobi in install documentation (this is an optional requirement)

### DIFF
--- a/discrete_optimization/rcpsp/solver/rcpsp_lp_solver.py
+++ b/discrete_optimization/rcpsp/solver/rcpsp_lp_solver.py
@@ -738,7 +738,7 @@ class LP_MRCPSP_GUROBI(GurobiMilpSolver, _BaseLP_MRCPSP):
             self.starts[task].start = self.start_solution.rcpsp_schedule[task][
                 "start_time"
             ]
-            self.model.addConstr(
+            self.model.addLConstr(
                 gurobi.quicksum(
                     [self.x[key] * key[2] for key in variable_per_task[task]]
                 )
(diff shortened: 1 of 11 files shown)


discrete_optimization/rcpsp/solver/rcpsp_lp_solver.py
```diff
@@ -666,7 +666,7 @@ def init_model(self, **args):
                 self.x[key] * key[2] for key in variable_per_task[last_task]
             )
         )
-        self.model.addConstrs(
+        self.model.addLConstrs(
```
g-poveda, line 669: to reverse (addLConstrs doesn't exist)

discrete_optimization/rcpsp/solver/rcpsp_lp_solver.py
```diff
@@ -685,7 +685,7 @@ def init_model(self, **args):
                 r: [non_renewable[r]] * len(self.index_time) for r in non_renewable
             }
 
-        self.model.addConstrs(
+        self.model.addLConstrs(
```
g-poveda, line 688: here too (and so on)

g-poveda (changes requested):
we have to keep addContr**s** as it was, and just replace the addConstr by addLConstr
otherwise it seems ok

nhuet:
> we have to keep addContr**s** as it was, and just replace the addConstr by addLConstr otherwise it seems ok

Done. (but already squashed in first commit, sorry about that)

g-poveda (approved):

PR #114, opened 2023-02-20
Set "trigger binder env build" step as optional in github workflow

nhuet:
Even if the build on binderhub servers failed to be launch for some reason (as network issues), it should not mark the whole job creating a new binder env as failed. As the binder environment as already been updated by first steps and the only drawback will be that the first user to click on a binder link will experiment a long waiting time before its notebook actually launches.


g-poveda (approved):
ok lgtm, when some cloud is down we still do the remaining actions.

PR #119, opened 2023-02-21
Use Agg backend for matplotlib in github actions to avoid random issues in windows tests

nhuet:
Tests involving matplotlib plots randomly fail on windows with tkinter backend (e.g. that [workflow](https://github.com/airbus/discrete-optimization/actions/runs/4234233092/attempts/1)) . As suggested [here](https://discourse.matplotlib.org/t/azure-pipeline-is-failing-randomly-when-using-plt-subplots-because-of-tcl/22893/2), we use agg backend instead to avoid that.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -227,7 +227,7 @@ jobs:
           # show library path used
           pytest -s tests/show_do_path.py
           # run test suite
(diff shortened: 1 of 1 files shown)


g-poveda:
Ok, hopefully we don't get anymore random crashes !

PR #127, opened 2023-05-15
Refactor non preemptive rcpsp models and fix typing issues

nhuet:
- [x] In the same way as it is handled in rcpsp_multiskill, RCPSPModel gather now single mode, multi mode, calendar, and special constraints models. The different characteristics are still available thanks to methods is_rcpsp_multimode(), is_varying_resource(), and includes_special_constraint().
- [x]  Similarly we merge RCPSPSolution and RCPSPSolutionSpecial.
- [x]  We split rcpsp_model.py into several pieces to avoid a much too long file
- [x]  We gather utility functions in rcpsp_utils.py
- [x]  We avoid some circular imports by importing classes for annotations only when type checking (see https://adamj.eu/tech/2021/05/13/python-type-hints-how-to-fix-circular-imports/)
- [x]  We fully annotate files and correct related typing issues, except for solvers and modules involving preemptive models .
- [x]  We introduce SolverRCPSP and SolverGenericRCPSP to help annotate rcpsp_solvers
- [x]  We move GHH and LS_RCPSP_Solver in generic_rcpsp_tools package as it can also apply to multiskill_rcpsp


discrete_optimization/generic_rcpsp_tools/generic_rcpsp_solver.py
g-poveda, line 1: For localsearch and large neighborhood search i guess, seems good.

PR #142, opened 2024-01-11
New solver for RCPSP models

g-poveda:
- Add a new solver based on ortools-cpsat solver. The model works for multimode-rcpsp, including the variable calendar, non renewable resource.


nhuet (approved):
I think the 3 commits can be squashed together as the 2 last ones are only corrections of the first one. Else LGTM!

PR #144, opened 2024-01-19
Update ortools solvers

nhuet:
- upgrade ortools dependency to >=9.8
- upgrade python minimal version to >=3.8 (ortools 9.8 does not exist for python<3.8 and python 3.7 is not maintained anymore)
- update knapsack solver based on ortools to adapt to the api changes
- update pickup_vrp solver based on ortools
  - construct firstsolutionstrategy and localsearchmetaheuristic enum from corresponding ortools pseudo enum
  - construct mapping from integer values to ortools solver status from ortools code
  - adapt tests and plot functions to work directly with ResultStorage and GPDPSolution instead of specific tuple
  - merge solve_intern into solve and construct directly resultstorage on the fly (could be useful to connect logging d-o callbacks or callbacks for hyperparamaters tuning with optuna)
  - fix ResultStorage.add_solution (needed by the previous item)

### DIFF
--- a/discrete_optimization/generic_tools/result_storage/result_storage.py
+++ b/discrete_optimization/generic_tools/result_storage/result_storage.py
@@ -64,32 +64,47 @@ class ResultStorage:
             if self.best_solution is None:
                 f = min if not self.maximize else max
                 self.best_solution = f(self.list_solution_fits, key=lambda x: x[1])[0]
+        else:
+            self.min = None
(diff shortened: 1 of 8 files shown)


g-poveda (approved):
Look good to me, worth to note that if in practice the lack of autocompletion when using the ortools "enum" reduce too much the easiness of development we may want to add some helper functionnality.

PR #145, opened 2024-01-22
Increase minimal version of python to >=3.8

nhuet:
- latest versions of dependencies need python>=3.8
- python 3.7 is not maintained anymore
- we test now for python 3.8 and 3.11


g-poveda (approved):
lgtm

PR #146, opened 2024-01-24
Adding VRP examples and test + update routing examples

g-poveda:
- Some standalone example script for the vrp problem are implemented
- unit test for ortools and greedy vrp solver implemented
- Also took this PR to update the example of the pickup_vrp problems, closely related to classical VRP, so that they run well with latest change in the library.
- fix LP solver for VRP

### DIFF
--- a/discrete_optimization/vrp/solver/lp_vrp_iterative.py
+++ b/discrete_optimization/vrp/solver/lp_vrp_iterative.py
@@ -476,19 +476,16 @@ class VRPIterativeLP(SolverVrp):
     def __init__(
         self,
         vrp_model: VrpProblem2D,
-        params_objective_function: ParamsObjectiveFunction,
+        params_objective_function: Optional[ParamsObjectiveFunction] = None,
     ):
-        SolverVrp.__init__(self, vrp_model=vrp_model)
+        SolverVrp.__init__(
+            self,
+            vrp_model=vrp_model,
+            params_objective_function=params_objective_function,
+        )
         self.model: Optional[Model] = None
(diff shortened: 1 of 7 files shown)


discrete_optimization/vrp/solver/lp_vrp_iterative.py
nhuet, line 1: It is not a big deal, but when you are modifying discrete_optimization/vrp/solver/*, you are not updating/adding tests and examples but rather fixing issues. 
Ideally you should split this commit into code fixing and tests/examples updating. Or you could add a line in your commit about fixing code.

examples/pickup_vrp/loading_example.py
```diff
@@ -230,9 +218,10 @@ def check_solution(res, gpdp: GPDP):
         n_solutions=10000,
     )
     logging.basicConfig(level=logging.DEBUG)
-    results = solver.solve_intern()
```
nhuet, line 233: I think there are other functions in this file that are using the old api `solve_intern()`. As they are not called in `if __name__ == "__main__": `block, the example is still working though. I made the modifications in my branch on ortools solver, so maybe you can let it as is and i will push my own modifications later.

nhuet (approved):

examples/pickup_vrp/loading_example.py
g-poveda: done

discrete_optimization/vrp/solver/lp_vrp_iterative.py
g-poveda: done, thanks !

PR #148, opened 2024-01-25
Add callbacks in `SolverDO.solve()`

nhuet:
We create a Callback class to be used by solver.solve() to hook at some stages.

This will enable
- logging to tools like Tensorboard or Optuna
- timing out solvers
- early stopping
- discarding not promising trials during hyperparameters tuning

Callback has a few methods to overwrite in order to execute some code
- `on_solve_start()`: at the beginning of solve
- `on_solve_end()`: at the end of solve
- `on_step_end()`: at the end of each iteration

We are deeply inspired by work from others:

- keras: https://github.com/keras-team/keras/blob/v3.0.2/keras/callbacks/callback.py
- tensorflow: https://github.com/tensorflow/tensorflow/blob/cafa8f3bd9b5740d1d37db799eec3168edc7e97b/tensorflow/python/training/session_run_hook.py#L94
- or-tools: https://developers.google.com/optimization/reference/python/sat/python/cp_model#cp_model.CpSolverSolutionCallback
- sklearn-genetic-opt: https://github.com/rodrigo-arenas/Sklearn-genetic-opt/blob/0.10.1/sklearn_genetic/callbacks/base.py

In this PR, we only make local search algorithms use them, we will add gradually this possibility to all solvers.
We replace some features (logging, backup, and timer) previously used by local search algos by the use of some basic callbacks that we implement, as shown in test tests/generic_tools/callbacks/test_sa_with_callbacks.py.

### DIFF
--- a/discrete_optimization/generic_rcpsp_tools/ls_solver.py
+++ b/discrete_optimization/generic_rcpsp_tools/ls_solver.py
@@ -119,7 +119,5 @@ class LS_RCPSP_Solver(SolverGenericRCPSP):
         result_sa = ls.solve(
             dummy,
             nb_iteration_max=kwargs.get("nb_iteration_max", 2000),
-            max_time_seconds=kwargs.get("max_time_seconds", None),
-            pickle_result=False,
         )
         return result_sa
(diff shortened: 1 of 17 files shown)


g-poveda (approved):
this looks excellent/promising for the future !

PR #154, opened 2024-02-27
Add SolverDO.__init__() and ._init_model()

nhuet:
It allows to put some boilerplate code in one place:
- set self.problem (and all attributes in child solvers are renamed into problem: rcps_model, coloring_model, ...)
- generate self.aggreg_from_sol, self.aggreg_from_dict, and self.params_objective_function (we unify also their names)

We add init_model() because functions in coloring.coloring_solvers, facility.facility_solvers, ... assume its existence to solve a problem from a solver class.

In a next step, these modules could also be made generic and put in common.

### DIFF
--- a/discrete_optimization/coloring/coloring_solvers.py
+++ b/discrete_optimization/coloring/coloring_solvers.py
@@ -108,19 +108,19 @@ def look_for_solver_class(
 
 
 def solve(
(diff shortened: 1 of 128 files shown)


g-poveda (changes requested):
Great thanks, need to update the api used in the notebook examples too.

PR #156, opened 2024-02-27
Adding cpsat solver for knapsack problem

g-poveda:
- adding example and test associated
- tested its integration in the meta solver called decomposition solver


g-poveda:
This PR has been included in #151

PR #161, opened 2024-03-05
Adding objective function logger callback

g-poveda:
-new callback that logs current best objective function -example of use in one example script

### DIFF
--- a/discrete_optimization/generic_tools/callbacks/loggers.py
+++ b/discrete_optimization/generic_tools/callbacks/loggers.py
@@ -49,3 +49,40 @@ class NbIterationTracker(Callback):
         logger.log(
             msg=f"Iteration #{self.nb_iteration}", level=self.step_verbosity_level
         )
+
+
+class ObjectiveLogger(Callback):
+    """
+    Stops the optimization process if a limit training time has been elapsed.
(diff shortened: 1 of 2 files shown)


discrete_optimization/generic_tools/callbacks/loggers.py
```diff
@@ -49,3 +49,40 @@ def on_step_end(
         logger.log(
             msg=f"Iteration #{self.nb_iteration}", level=self.step_verbosity_level
         )
+
+
+class ObjectiveLogger(Callback):
+    """
```
nhuet, line 55: Maybe the docstrings should be changed ;p

nhuet (changes requested):

discrete_optimization/generic_tools/callbacks/loggers.py
g-poveda: well seen, actually same for the already existing NbIterationLogger right ?!

nhuet: ... correct.

g-poveda: Done

PR #162, opened 2024-03-07
Put in common `MilpSolver.retrieve_solutions()`

nhuet:
Now milp solvers need rather to implement `retrieve_ith_solution()` as they already all loop in the same way to get all solutions.


g-poveda (approved):
Ok look good to me, but may change when adding the callback logic

PR #168, opened 2024-03-21
Implement callbacks integration for gurobi solvers

nhuet:
- replace method to implement (for all milp solvers) `retrieve_ith_solution()` by `retrieve_current_solution()` that take a callable  `get_var_value_for_current_solution()`. Then
  - `retrieve_ith_solution` is `retrieve_current_solution` with
    `get_var_value_for_ith_solution`
  - during a gurobi callback, it will be called with `model.cbGetSolution`
- update milp solvers to reflect that
- create a `GurobiCallback` class that
   - populates on the fly a result_storage by using
     `retreive_current_solution()`
   - calls user-defined d-o callbacks
   - terminate solve if user-defined callbacks decide to early stop
- add a test using d-o callbacks with gurobi for coloring solver
- the gurobi pickup-vrp solver still does not use callbacks as it is
  overriding `GurobiMilpSolver.solve()` to perform a loop.

### DIFF
--- a/discrete_optimization/coloring/solvers/coloring_lp_solvers.py
+++ b/discrete_optimization/coloring/solvers/coloring_lp_solvers.py
@@ -6,7 +6,7 @@
 
 import logging
 import sys
-from typing import Any, Dict, Hashable, Optional, Tuple, Union
+from typing import Any, Callable, Dict, Hashable, Optional, Tuple, Union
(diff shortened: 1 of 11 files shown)


g-poveda (approved):
Tested it on my side, look good to me ! thanks

PR #175, opened 2024-03-28
Update solvers to make them use `complete_with_default_hyperparameters`

nhuet:
The idea is
- to define once for all default values for hyperparameters when
  defining them in solver.hyperparameters
- reuse these default values to complete kwargs in solve/init_model in
  one line

Thus
- we add default values in hyperparmeters definitons when missing
-  we call `complete_with_default_hyperparameters` before getting
hyperparameter values from kwargs


g-poveda (approved):
Lgtm

PR #178, opened 2024-04-04
Update Knapsack solver and optuna examples

g-poveda:
- Running optuna_example helped identify minor problem on KnapsackDynProg solver (silent bug), and on retrieve solution of a CP solver
- update the optuna example script to run some more interesting experiment.

### DIFF
--- a/examples/knapsack/optuna_example.py
+++ b/examples/knapsack/optuna_example.py
@@ -26,18 +26,19 @@ from discrete_optimization.knapsack.knapsack_parser import (
     get_data_available,
     parse_file,
 )
-from discrete_optimization.knapsack.solvers.cp_solvers import CPKnapsackMZN2
+from discrete_optimization.knapsack.solvers.cp_solvers import (
+    CPKnapsackMZN,
+    CPKnapsackMZN2,
+)
 from discrete_optimization.knapsack.solvers.dyn_prog_knapsack import KnapsackDynProg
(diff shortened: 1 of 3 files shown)


examples/knapsack/optuna_example.py
```diff
@@ -155,20 +157,26 @@ def objective(trial: Trial):
...
     solver.init_model(**kwargs)
 
-    # solve
-    sol, fit = solver.solve(
-        callbacks=[
-            OptunaPruningSingleFitCallback(trial=trial, **kwargs),
-        ],
-        **kwargs,
-    ).get_best_solution_fit()
+    if True:
```
nhuet, line 160: Why this `if True: ... else: raise TrialPruned()` ?

examples/knapsack/optuna_example.py
```diff
@@ -155,20 +157,26 @@ def objective(trial: Trial):
...
+    else:
+        raise optuna.TrialPruned()
 
     return fit
 
 
 # create study + database to store it
 storage = JournalStorage(JournalFileStorage(storage_path))
-optuna.delete_study(study_name=study_name, storage=storage)
+try:
```
nhuet, line 176: Sorry, i should have removed this line myself, it was to reset a previous study. As it is now it starts from scratch at each script run. If you remove or comment these lines, it optimizes further the study, remembering the previous trials.

nhuet (approved):

PR #181, opened 2024-04-05
Implement callbacks for lns solvers

nhuet:
- solve() is an alias of solve_lns() with same arguments
- remove arg max_time_seconds beacause can be handled by callback TimerStopper


g-poveda (approved):
lgtm

PR #184, opened 2024-04-08
Fix CategoricalHyperparameter.suggest_with_optuna()

nhuet:
Was not taking into account choices arg

Also add kwargs to all suggest_with_optuna signature


g-poveda (approved):
thanks !

PR #187, opened 2024-04-09
Implement callbacks for minizinc-based solvers

nhuet:
We use the `instance.output_type` as an entry point.
During the `solve()` call, we create dynamically a subclass of `MinizincCPSolution` which:
- wraps an actual d-o solution
- has class attributes properly initialized for tracking
  - full `ResultStorage`
  - current number of step
  - solver instance
  - user-defined callbacks
This class will be called each time a new solution is found by minizinc, and thus update the result storage and call upon user-defined callbacks.

If a stopping criteria is reached in callbacks, we raise a `SolveEarlyStop` exception which will be caught in `solve()`.

In addition to that a `MinizincCPSolver` must now implement `retrieve_solution()`
which converts a solution from minzinc into a d-o one.
It replaces former `retrieve_solutions()` and means that solutions will be now converted on the fly, and directly added to the current result_storage.

We also fix other bugs in minizinc solvers on the way:
- when using latest version of minzinc, rcpsp_multiskill solver was raising an issue, we fixed the minizinc file defining the model
- we upgrade the minizinc version used by github runners to test compatibility of our solvers with latest version of minizinc
- we fix CP_MULTISCENARIO, which was not really filling the result storage with solutions but rather with tuple of solutions for different variations of the same model. Instead,we make it work on AggregRCPSPModel so that the fitness is computed correctly on the list of submodels.

### DIFF
--- a/.github/workflows/build-doc.yml
+++ b/.github/workflows/build-doc.yml
@@ -41,9 +41,10 @@ jobs:
       python-version: "3.9"
       minizinc_config_cmdline: export PATH=$PATH:$(pwd)/bin/squashfs-root/usr/bin; export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:$(pwd)/bin/squashfs-root/usr/lib
       minizinc_cache_path: $(pwd)/bin/squashfs-root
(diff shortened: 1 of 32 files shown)


g-poveda (approved):
👍

PR #190, opened 2024-04-11
Fix optuna callback

nhuet:
Optuna callback was relying on `ResultStorage.best_fit` to get the current best fitness. But this property is correctly filled only if  the result_storage has been filled using `add_solution()`. And lost of solvers actually directly append solutions to the attribute `ResultSorage.list_solution_fits`.

So we rather use `ResultStorage_get_best_solution_fit()` which works always.


g-poveda (approved):
Ok

PR #196, opened 2024-04-25
Modified parse_file function to include the possibility to read in patterson rcpsp files

derhendrik:
Included a function "parse_patterson" to read in RCPSP-instances in Patterson file format.
This file format is (by now) much more wide-spread than the explicit file format of PSPLIB-instances.

The "parse_patterson" method is called, if an RCPSP-instance-file ends with ".rcp".

See also here: https://www.projectmanagement.ugent.be/research/project_scheduling/rcpsp

e.g. datasets RG30 and RG300

### DIFF
--- a/discrete_optimization/rcpsp/rcpsp_parser.py
+++ b/discrete_optimization/rcpsp/rcpsp_parser.py
@@ -27,7 +27,9 @@ def get_data_available(
 
     try:
         files = [
-            f for f in os.listdir(data_folder) if f.endswith(".sm") or f.endswith(".mm")
+            f
+            for f in os.listdir(data_folder)
+            if f.endswith(".sm") or f.endswith(".mm") or f.endswith(".rcp")
         ]
     except FileNotFoundError:
         files = []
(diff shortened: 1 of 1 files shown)


g-poveda:
Thanks, nice idea :), i had in mind to include some RG instances so this parser will help. should be ok to merge soon.

g-poveda:
did you validate that the instances are correct and that let's say a basic solution can be computed with some solver in discrete-optimization ? 
on my side i think i would integrate the inclusion of RG30/300 in the datasets.py file and include a test based on 1 given file

derhendrik:
Hey! I did compute solutions with the CP-Solver (MiniZinc) which works well, however I have not made any further checks (manually checking resource and precedence feasibility of the generated solution). 

Please also note that I had to fix a small bug with my latest commit.
(Was difficult to see because RG30-instances have resource-availabilities of [10, 10, 10, 10] and I mostly tested on RG30 instances. Other instances, such as the sD-data set however may have unequal resource-availabilities such as [16, 12, 13, 15]).

g-poveda:
Hi again, yes thanks to have corrected your parser, i looked into it and tested also on sd data and it seems fine. 
We added tests and also added some automatic download of RCP datasets in the datasets.py script, in the next PR #197.
do you know if rcp format has been worked also for multimode rcpsp, i don't see it from a first look. 
Cheers

derhendrik:
Hey! As far as I know, there is no multi-mode .rcp files.
However, there is a set of these "Boctor"-files for the multi-mode case. (See here: https://www.om-db.wi.tum.de/psplib/dataob.html).

However, their file-ending seems to be ".prb". The files are based on this paper: "Heuristics for scheduling projects with resource restrictions and several resource-duration modes", however I did not look deeper into what sort of information is in these files.

PR #199, opened 2024-04-29
Fix github actions with macos-latest

nhuet:
The github runner "macos-latest" is now an aliasfor macos-14 on arm64.

- There is no gurobipy for python 3.8 +  macos arm64, so we use macos-12 (x86_64) instead when testing on python 3.8.
- We use prerelease 1.16 for mip when on macos-latest to avoid an error with cbc-mip, waiting for mip 1.16 to be officially released (see  https://github.com/coin-or/python-mip/issues/335#issuecomment-1986745529)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -156,6 +156,11 @@ jobs:
         with:
           name: dist
           path: dist
+      - name: Install prerelease version of pymip (only for macos arm64)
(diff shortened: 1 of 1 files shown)


g-poveda (approved):
Thanks ! let's not forget those (hopefully) temporary fixes.

PR #209, opened 2024-05-22
Add maximum independent set problem and solvers

ArmandGautier:
Add few solvers to solve the maximum independent set problem, like gurobi, ortools, kamis and two solvers using hybrid quantum algorithm with qiskit.


g-poveda:
unit test with this code :
small_example = get_data_available()[0]
could be made more robust so that we pick always the same file from one computer to another one (so maybe picking the right filename instead of taking the [0] element. I think it can explain some fail of unit test @ArmandGautier

g-poveda (approved):
lgtm

PR #213, opened 2024-05-27
Skip tests using python-mip on macos arm64

nhuet:
There are issues with cbc included in the wheel for this platform. See https://github.com/coin-or/python-mip/issues/167


g-poveda (approved):
thx !

PR #216, opened 2024-05-30
add optuna example and two new solvers for mis problem

ArmandGautier:
Add optuna example to test mis solvers with the optuna tool. Add two new solvers : one unsing a networkX method and another one using a quadratic formulation with gurobi.


g-poveda (approved):
ok !

PR #217, opened 2024-06-03
Allow SubBrickKwargsHyperparameter to directly specify a subbrick class

nhuet:
When only one class is possible, instead of having a dedicated hyperparameter with only 1 choice, we directly specify the class in the SubBrickKwargsHyperparameter.

We implement that with GA_RCPSP_Solver and ParametersGa.


nhuet:
Split this PR into 2 PRs.

PR #218, opened 2024-06-04
Allow SubBrickKwargsHyperparameter to directly specify a subbrick class

nhuet:
When only one class is possible, instead of having a dedicated hyperparameter with only 1 choice, we directly specify the class in the SubBrickKwargsHyperparameter.


g-poveda (approved):
great

PR #221, opened 2024-06-06
Fix issue with subsolvers whose hyperparameters share names but not types

nhuet:
Optuna suggestion could raise error when a subbrick could take 2 different classes that have both an hyperparameter with same name but of a different kind. (For instance one being an `IntegerHyperparameter`, the other being a `CategoricalHyperparameter`).

To avoid this issue, we update the hyperparameter prefix when seen as a parameter of an optuna trial to include the actual class name of the chosen subbrick.

For example,

```python
class MySolver(SolverDO):
    hyperparameters = [
        SubBrickHyperparameter("subsolver", choices=[LPSolver, CPSolver]),
        SubBrickKwargsHyperparameter(
            "kwargs_subsolver", subbrick_hyperparameter="subsolver"
        ),
    ]

    ...

class LPSolver(SolverDO):
    hyperparameters = [
        CategoricalHyperparameter("hp", choices=[True, False]),
    ]
    ...

class CPSolver(SolverDO):
    hyperparameters = [
        IntegerHyperparameter("hp", low=0, high=10),
        FloatHyperparameter("coeff", low=0., high=1.),
    ]
    ...
```

Here the trial params will be

- first trial: `(subsolver, subsolver.LPSolver.hp)`
- second trial: `(subsolver, subsolver.CPSolver.hp, subsolver.CPSolver.coeff)`

We modify `DummySolver2` in test_hyperparameters.py so that the issue raised in the tests before the fix and the tests are ok with it.

NB: The issue could still arise if a subbrick could be of two different classes with same name (but from different modules) that also share same hyperparameter name of different type. This is higly unlikely to happen and could not justify having the whole path including the module to be inserted in the prefix. This would lead to a very unreadable parameter name in optuna trial.

### DIFF
--- a/discrete_optimization/generic_tools/hyperparameters/hyperparametrizable.py
+++ b/discrete_optimization/generic_tools/hyperparameters/hyperparametrizable.py
@@ -248,9 +248,10 @@ class Hyperparametrizable:
                     ] = suggested_and_fixed_hyperparameters[
                         hyperparameter.subbrick_hyperparameter
                     ]
-                    kwargs_for_optuna_suggestion[
-                        "prefix"
-                    ] = f"{prefix}{hyperparameter.subbrick_hyperparameter}."
+                    kwargs_for_optuna_suggestion["prefix"] = (
+                        f"{prefix}{hyperparameter.subbrick_hyperparameter}."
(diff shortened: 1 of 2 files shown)


g-poveda (approved):
lgtm thx

PR #224, opened 2024-06-06
Remove unused imports

nhuet:
We pass the command

```shell
autoflake -r  --remove-all-unused-imports --in-place  discrete_optimization
```

And then revert some changes to keep import like :

```python
try:
    import gurobipy
except ImportError:
    gurobi_available = False
else:
    gurobi_available = True
    from gurobipy import GRB, LinExpr, Model
```

And change imports when it was not from the direct source: `from discrete_optimization.generic_tools.lns_cp import TrivialPostProcessSolution` -> `from discrete_optimization.generic_tools.lns_tools import TrivialPostProcessSolution`

### DIFF
--- a/discrete_optimization/generic_tools/lp_tools.py
+++ b/discrete_optimization/generic_tools/lp_tools.py
@@ -33,9 +33,6 @@ else:
 
 try:
     import docplex
-    from docplex.mp.constr import LinearConstraint
-    from docplex.mp.dvar import Var
-    from docplex.mp.model import Model
     from docplex.mp.progress import SolutionListener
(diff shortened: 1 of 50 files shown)


discrete_optimization/generic_tools/lp_tools.py
```diff
@@ -33,9 +33,6 @@
 
 try:
     import docplex
-    from docplex.mp.constr import LinearConstraint
```
g-poveda, line 36: to be checked if it doesn't create problem

g-poveda (changes requested):

PR #225, opened 2024-06-10
Update ParamsConstraintBuilder hyperparameters

nhuet:
- add a "high" value for each hyperparameter (can still be tuned in a specific study)
- when their are used as subbrick as only class, use subbrick_cls to avoid having a useless subbrick hyperparameter always set to ParamsConstraintBuilder


g-poveda (approved):
Ok

PR #226, opened 2024-06-10
add a step parameter to the class IntegerHyperparameter

ArmandGautier:
add a step parameter to the class IntegerHyperparameter

### DIFF
--- a/discrete_optimization/generic_tools/hyperparameters/hyperparameter.py
+++ b/discrete_optimization/generic_tools/hyperparameters/hyperparameter.py
@@ -101,6 +101,8 @@ class IntegerHyperparameter(Hyperparameter):
     If None, the hyperparameter value has no upper bound.
 
     """
+    step: int = 1
+    """step to discretize."""
 
     default: Optional[int] = None
     """Default value for the hyperparameter.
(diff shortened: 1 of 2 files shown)


discrete_optimization/generic_tools/hyperparameters/hyperparameter.py
```diff
@@ -100,6 +100,12 @@ class IntegerHyperparameter(Hyperparameter):
 
     If None, the hyperparameter value has no upper bound.
 
+    """
+    step: Optional[int] = 1
+    """step to discretize.
+
+    If None, the step will be one.
+
     """
```
nhuet, line 109: Write rather
```
step: int = 1
"""Step to discretize."""
```

This is sufficient to have step default value being 1, and enforce an integer value.
Actually step=None would make trial.suggest_int() fail.

nhuet (changes requested):
A unit test should be using this new parameter. 
For instance, in tests/generic_tools/hyperparameters/test_hyperparameter.py, update the hyperparameter "nb" of "MetaSolverFixedSubsolver" with step=2 and high=2 (so that the number of cases found by BruteForceSampler is still the same)

PR #228, opened 2024-06-10
Replace test for timed pruner with timed optuna callback

nhuet:
Instead of using SA which can sometimes result in no pruning depending on the platform, we use an ad-hoc solver which always returns the same sequence of intermediate values with only the speed varying (via a controlled `sleep` call)

=> with timed pruner, pruning occurs
=> with classic pruner, pruning does not occur

Moreover each test takes only 2 seconds as we choose the speed of the solver.


g-poveda (approved):
thanks way better !

PR #238, opened 2024-06-17
Prepare compatibility with numpy 2.0

nhuet:
cf https://numpy.org/devdocs/numpy_2_0_migration_guide.html

float_ -> float64


g-poveda (approved):
seems fine !

PR #249, opened 2024-07-05
Implement lexicographic optimization

nhuet:
- Add a `LexicoSolver` which makes a given subsolver solve sequentially
  several specified objectives, while adding a constraint at each step
  to avoid worsening results according to previous objectives

- Add a bunch of methods to implement (with partial default
  implementation) so that a solver can be used as a subsolver by
  `LexicoSolver`:
  - `implements_lexico_api()`: returns `True` if the solver is lexico ready
    (ie implement subsequent methods). `LexicoSolver` raise a warning if
    this is not the case
  - `get_model_objectives_available()`: list of labels available
    corresponding to the intern objectives the solver can optimize.
    Defaults to keys of the problem's `ObjectiveRegister` (via new method
    `problem.get_objective_names()`)
  - `set_model_objective()`: update the intern objective the subsolver
    will optimize
  - `get_model_objective_value()`: retrieve the value of the intern
    objective currently optimized
  - `add_model_constraint()`: add a constraint to the intern model
    on the given objective to avoid worsening it.

- Implement them for ortools-csat solvers on knapsack and rcpsp

### DIFF
--- a/discrete_optimization/generic_tools/do_problem.py
+++ b/discrete_optimization/generic_tools/do_problem.py
@@ -213,6 +213,9 @@ class ObjectiveRegister:
         self.objective_handling = objective_handling
         self.dict_objective_to_doc = dict_objective_to_doc
 
+    def get_objective_names(self) -> List[str]:
+        return sorted(self.dict_objective_to_doc)
+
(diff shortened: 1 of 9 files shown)


g-poveda (approved):
lgtm

PR #255, opened 2024-07-11
add general class solvers for QAOA and VQE, can be use for any DO problem with a gurobi model. + quantum solvers for the knapsack problem

ArmandGautier:


g-poveda (approved):
lgtm

PR #260, opened 2024-07-29
Refactor SubBrickHyperparameter

nhuet:
- previous SubBrickHyperparameter is renamed SubBrickClsHyperparameter
- new SubBrickHyperparameter directly suggests class *and* kwargs of a  sub-brick. We cannot suggest the Hypeparametrizable instance as we need kwargs to use it also in init_model() and solve() for instance so we rather suggest a SubBrick instance (new class)
- SubBrick class = wrapper to hold class and kwargs of a   Hyperparametrizable object


g-poveda (approved):
excellent thanks

PR #264, opened 2024-07-30
Add WarmstartMixin for solvers which can start from an initial solution

nhuet:
- Add a mixin class to inheritate from when a warm start is possible for a solver
- Implement set_warm_start for 
    - ortools-cpsat
    - lns
    - gurobi
    - ortools-routing
    - decomposed solvers
    - local search
    - genetic algorithms
 - Add a test for each implementation


g-poveda (approved):
lgtm

PR #267, opened 2024-08-02
Fix bug in cpsat vrp model

g-poveda:
1 constraint was incorrect and was not insuring that when a given vehicle was used it was starting from the starting node.
It had the effect to give produce (sometimes) empty path in the retrieve solution. This should be fixed now.
- better unit test to check if the issue remains.


nhuet (approved):
lgtm.

PR #269, opened 2024-08-09
Implement a sequential metasolver

nhuet:
- Chain several subsolvers
- Next subsolver warm started by the best solution of the previous one
- subsolvers must inherit from `WarmstartMixin`, except for first one
- `**kwargs` needed in all solvers `__init__()` method as we apply
   - `subsolver.__init__(problem=problem, **kwargs)`
   - `subsolver.init_model(**kwargs)`
   - `subsolver.solve(**kwargs)`


g-poveda (approved):
lgtm

PR #272, opened 2024-08-13
List without replacement of categorical or subbrick hyperparameters 

nhuet:
We implement child classes of ListHyperparameter that does not allow
multiple of the same value in the suggested list.

We cannot restrict dynamically the choices of the hyperparameter
because optuna complains if a distribution change.
And the second hyperparameter generated (say hp_1) would have its
distribution modified according to the first one (hp_0).
Ex:  hp_template.choices = A, B, C
     if hp_0's value is  A, then hp_1.choices = B,C
     if                  B,                     A,C

So the solution chosen here is to raise a (custom) TrialDropped exception if the
list suggested is illicit (with duplicates), and to add this type of
exception to `catch` argument of `study.optimize()` to carry on with the study.
This is already done in the studies generated by `generic_optuna_experiment_xxx()`.


g-poveda (approved):
lgtm

PR #278, opened 2024-09-09
Harmonize time_limit/timeout/... arguments

nhuet:
Now we use always the arg `time_limit` instead of other names. We pout it out from `parameters_cp` and `parameters_milp`, so that any solver with a settable time_limit has a solve argument called "time_limit".
`time_limit_iter0` is also out of `parameters_cp` and becomes another arg of `BASE_LNS_CP.solve()`.


g-poveda (approved):
lgtm, warning API breaks on next release :)

PR #280, opened 2024-09-12
Add a tutorial for lexicographic optimization

nhuet:


g-poveda:
Great notebook, maybe we could think of a "multiobjective" logger in the future, because here in the second step of the optim, we don't see the current objective optim being optimized, we just see it via the callback called at the end of the subsolver call. that's a minor comment though

PR #286, opened 2024-09-16
Test on python 3.12

nhuet:
- Enforce cffi>=1.17 (updated for python 3.12, avoid weird errors for dependencies relying on C code)
- Update time limits to avoid errors with ubuntu + python 12: It seems ortools gets slower in its latest release 9.11 on ubuntu...
- Fix LP_MRCPSP_GUROBI for python 3.12: we have a crash when using gurobi.Model.addConstrs with a quicksum on a list rather than on a generator. In other words, we need to get rid of []. See https://support.gurobi.com/hc/en-us/community/posts/28192464443281-model-addConstrs-breaking-sometimes-after-Python-version-change for more details.
- Add a test dedicated to LP_MRCPSP_GUROBI.


g-poveda:
Interesting, in the link you provide they suggest that they no longer advise to use addConstrs, it seems they prefer something more manual it's fine to me, maybe we'll do that in the future

PR #292, opened 2024-09-19
Drop support for python 3.8

nhuet:
Why?
- python 3.8 is entering  end-of-life status (https://devguide.python.org/versions/#status-of-python-versions)
- ortools.mathopt is not supporting python 3.8 (as using new annotations style)

We take the opportunity to use >=3.9 type annotations style (List[] -> list[], ..., use Iterable, Sequence,... from collections.abc).


g-poveda (approved):
lgtm, warning big pr :)

PR #293, opened 2024-09-23
Fix quantum solvers `__init__()` and `init_model()`

nhuet:
They should all accept` **kwargs` and use it properly (`**` missing sometimes).
Generic solve functions (as used in optuna study) will try to pass kwargs to `__init__`, `init_model` and `solve` methods.


g-poveda (approved):
lgtm

PR #294, opened 2024-09-23
CPSat solver for multiskill rcpsp

g-poveda:
First basic native cpsat model for multiskill problem.
- the cpsat model should be quite flexible to different variants of multiskill model :
       - variants with 1 worker per task
       - possibly only 1 skill used per worker for a given task (which is the most used mcpsp in the litterature but not really the one coded so far in do)

### DIFF
--- a/examples/rcpsp_multiskill/mslib/mslib_parse_and_solve.py
+++ b/examples/rcpsp_multiskill/mslib/mslib_parse_and_solve.py
@@ -1,22 +1,28 @@
-#  Copyright (c) 2022 AIRBUS and its affiliates.
+#  Copyright (c) 2024 AIRBUS and its affiliates.
 #  This source code is licensed under the MIT license found in the
 #  LICENSE file in the root directory of this source tree.
 import logging
 import os
 
 from discrete_optimization.generic_rcpsp_tools.ls_solver import LS_RCPSP_Solver
+from discrete_optimization.generic_tools.callbacks.loggers import ObjectiveLogger
+from discrete_optimization.generic_tools.cp_tools import ParametersCP
 from discrete_optimization.rcpsp_multiskill.plots.plot_solution import (
     plot_resource_individual_gantt,
     plt,
 )
+from discrete_optimization.rcpsp_multiskill.rcpsp_multiskill import MS_RCPSPSolution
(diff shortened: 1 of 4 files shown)


examples/rcpsp_multiskill/mslib/mslib_parse_and_solve.py
```diff
@@ -32,5 +38,38 @@ def example_parsing_and_local_search():
...
+    files_dict = get_data_available()
+    file = [f for f in files_dict["MSLIB4"] if "MSLIB_Set4_1003.msrcp" in f][0]
+    model = parse_file_mslib(file, skill_level_version=False)
+    cp_model = CPSatMSRCPSPSolver(
+        problem=model,
+    )
+    cp_model.init_model(
+        one_worker_per_task=False, slack_skill=False, one_skill_per_task=True
+    )
+    cp_model.cp_model.Minimize(cp_model.variables["makespan"])
```
nhuet, line 51: Why this line? This already done in cp_model.init_model() I think ?
And if the goal was to show how to change the objective, why not use rather the higher level api: cp_model.set_model_objective("makespan")?

tests/rcpsp_multiskill/test_mspsp_imopse_cpsat.py
```diff
@@ -0,0 +1,32 @@
...
+def test_imopse_cpsat():
+    file = [f for f in get_data_available() if "100_5_64_9.def" in f][0]
+    model, _ = parse_file(file, max_horizon=1000)
+    cp_model = CPSatMSRCPSPSolver(
+        problem=model,
+    )
+    cp_model.init_model(
+        one_worker_per_task=True,
+    )
+    cp_model.cp_model.Minimize(cp_model.variables["makespan"])
```
nhuet, line 28: Same. Why this line?

nhuet (approved):
Apart from the 2 lines resetting the objective outside init_model(), lgtm.

tests/rcpsp_multiskill/test_mspsp_imopse_cpsat.py
g-poveda: it's not needed anymore but i think i was testing other objectives ^^, will maybe remove

PR #298, opened 2024-09-24
Unify set_warm_start for lp solvers (gurobi and ortools.mathopt)

nhuet:


g-poveda (commented):
I am not sure i totally got the philosophy of this PR, the method set_warm_start is not totally shared by the gurobi and mathopt solver, i need more explanation later.

PR #299, opened 2024-09-24
Bugs in rcpsp multiskill calendars utils and cpsat solver

g-poveda:
- in this PR we insure in a more robust way that the shape of vector we add together are coherent.
- this solved normally the failed actions appearing on PR #296


g-poveda (commented):
the failed action is not related to the fix it seems, and the other seem green so let's go

PR #305, opened 2024-09-30
Unify lp solvers `init_model()` methods 

nhuet:
Share as much as possible init_model() between gurobi and mathopt solvers, by using new methods helping creating model, variables and constraints for each framework.

We also added tests for init_model options that were not tested (like `use_cliques` and `greedy_start`).


g-poveda (approved):
lgtm !

PR #311, opened 2024-10-01
Stop checking minzinc binary at discrete_optimization import

nhuet:
Instead, we raise the same error (with explicit message) only when initializing a minizinc-based solver, thanks to the method `find_right_minizinc_solver_name()` used by each solver.

We transform the relevant test in github workflow accordingly, and update the docs.

We also upgrade the minimal requirement to 2.8.


g-poveda (approved):
Lgtm

PR #315, opened 2024-10-03
Update lexico api

nhuet:
- we want to make it clear that it deals with lexico optimization (e.g. `add_model_constraint` was not very clear)
- we want to avoid confusion with utility methods to be added to share most `init_model()` code for lp or cp solvers (in particular `set_model_objective` and `add_model_constraint`)

- we also replace "intern" by "internal" as a more appropriate english word. This was in particular used in docstring of lexico methods.


g-poveda (approved):
lgtm

PR #320, opened 2024-10-08
Adapt pickup-vrp lp iterative solver with ortools/mathopt

nhuet:
- We put as much code in common as possible
   - between gurobi and mathopt version
   - between normal/lazy/cluster versions
- We catch on the fly solutions as TemporaryResult during solve_one_iteration() thanks to a dedicated callback
- We put much part of solve() for base classes GurobiMilpSolver and OrtoolsMathOptMilpSolver into optimize_model() in order to make a good use of it (in particular callback stuff) in solve_one_iteration()
- We add some methods to base classes GurobiMilpSolver and OrtoolsMathOptMilpSolver:
  - set_random_seed()
  - add_linear_constraint_with_indicator() which mimics gurobipy.Model.addGenConstrIndicator()
- We add access to objective during MathOptCallback (by evaluating registered objective): Be careful, by default we use `as_linear_expression`. One has to override `has_quadratic_objective` to make it use `as_quadratic_expression` instead.
- We remove dead code (if do_order in update_model(), init_warm_start())

### DIFF
--- a/discrete_optimization/generic_tools/lns_mip.py
+++ b/discrete_optimization/generic_tools/lns_mip.py
@@ -37,8 +37,7 @@ class GurobiConstraintHandler(ConstraintHandler):
         previous_constraints: Iterable[Any],
         **kwargs: Any,
     ) -> None:
-        solver.model.remove(list(previous_constraints))
-        solver.model.update()
+        solver.remove_constraints(previous_constraints)
 
 
 class OrtoolsMathOptConstraintHandler(ConstraintHandler):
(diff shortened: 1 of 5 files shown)


g-poveda (commented):
for i in range(self.nb_solutions - 1, -1, -1):
            list_temporary_results.append(self.retrieve_ith_temporaryresult(i=i))
 Line 1458, is this function still existing ?

nhuet:
> for i in range(self.nb_solutions - 1, -1, -1): list_temporary_results.append(self.retrieve_ith_temporaryresult(i=i)) Line 1458, is this function still existing ?

The lazy version of the solver is never tested and this is an issue. I will write a dedicated test and ensure it is working.

g-poveda (approved):
lgtm

PR #328, opened 2024-10-14
Full renaming (package/module/class)

nhuet:
The goal is to have more consistency in names choices across the
library, between and even within problem type.

- We use camel case for classes
- For each problem,
   - the solvers are stored in a subpackage "solvers" (was sometimes "solver"),
     in a module with a short name "{technology}" without adding
     "solver" or the problem name
   - the base solver class is also in solvers, accessible directly via
     `discrete_optimization.{problem}.solvers`,
   - the solvers_map is accessible in `discrete_optimization.{problem}.solvers_map`,
   - the solver class name follow the logic `{Technology}{Problem}Solver`,
   - the problem class name is `{Problem}Problem`,
   - the solution class name is `{Problem}Solution`,
- To be consistent, problem names which are acronym keeps only their
  first letter in uppercase: Rcpsp, Fjsp, Jsp, Gpdp, Vrp, Tsp
- pickup_vrp becomes Gpdp (was already used for problem, solution and
  some solvers)
- all CP (resp. LP) becomes generally Cp (resp. Lp), to avoid confusion
  between (formerly) CPM_RCPSP and CP_MRCPSP.
- MRCPSP -> MultimodeRcpsp
- RCPSP_MS -> MultiskillRcpsp

For instance the cp-sat solver for coloring can be imported via
```
from discrete_optimization.coloring.solvers.cpsat import CpSatColoringSolver
```
and the problem class via
```
from discrete_optimization.coloring.problem import ColoringProblem
```

- tests and examples names are also harmonized (less prefix/suffix)


g-poveda (approved):
thanks for the big work

PR #335, opened 2024-10-17
Fix lns init phase + constraint handler mix for minizinc subsolvers

nhuet:
- Fix ConstraintHandlerMix for minizinc subsolvers (and add test)
- Raise errors when lns has no initial_solution_provider but does not skip the initial phase or when it does not return a solution


g-poveda (approved):
lgtm

PR #339, opened 2024-10-17
Wrap (time, node) in CPM so that heapq works with mixed node types

nhuet:
We wrap it in a class that compares only on time so that heapq do not care if several node types occur.
(Was an issue in examples.jsp.run_dp.run_dp_of_rcpsp)


g-poveda (approved):
lgtm

PR #344, opened 2024-10-18
Update rcpsp + lns tutorial

nhuet:
- the constraint-handler specific to rcpsp does not exist anymore
- => we skip the part using the base lns solver
- we adapt the arguments used by LnsCpMznGenericRcpspSolver
  - remove or rename not relevant parameters


g-poveda (approved):
lgtm

PR #363, opened 2025-04-25
Introduce a generic cpmpy solver

nhuet:
- Derived in knapsack
- Has a dedicated method to explain unsatisfiability


g-poveda:
Ok this is excellent, i think we'll work next on notebook/tuto and ways of visualising the metaconstraint for some example : coloring seems to be a good candidate

PR #364, opened 2025-04-25
Implement explain_unsat() for gurobi solvers

nhuet:


g-poveda (approved):
ok lgtm, later we might want to use also metaConstraint for milp based solver.

PR #366, opened 2025-05-20
Fix internal objective and bound for cpsat

nhuet:
- when no sol found (e.g. timeout) => we want to get None instead of 0 (which could mean optimal value found)
- when optimal, we should have bound == obj. We need for that to call `self.solver.BestObjectiveBound()` after solve, which is tighter than the last call to `self.clb.BestObjectiveBound()`


g-poveda (approved):
thanks, better this way

PR #367, opened 2025-05-20
Fix current best objective value and bound for gurobi and mathopt solvers

nhuet:
- Use `gurobipy.GRB.Callback.MIPSOL_OBJ` to track current best objective value instead of
`gurobipy.GRB.Callback.MIPSOL_OBJBST` which seems not to take into account last sol (and thus provides a "infinite" one when called on the first solution found)
- Update obj and bnd at the end of the solve process (intermediate results can be not tight)


g-poveda (approved):
lgtm

PR #372, opened 2025-05-27
Add doc about dashboard

nhuet:


docs/source/dashboard.md
```diff
@@ -0,0 +1,141 @@
...
+```
+
+
+## Launching a d-o study
+
+We call here a study a set of experiments, where each experiment is the choice of
+- a solver config (class and hyperparameters)
+- a problem instance to solve
+
+### Experiment
```
g-poveda, line 34: for experiment and stats retriever callbacks, i would be interested to get code snippets examples, actually showing how to launch a study.
code snippet a bit like you put on other subsections.

g-poveda (changes requested):
If possible add more code snippet for the section i identified. except that, all good !!

PR #374, opened 2025-06-12
Update explainability methods with meta constraints

nhuet:
- current methods that deduce meta constraints set from minimal fine constraints set are renamed "`***_deduced_meta`" instead of "`***_meta`".
- to find a minimal of meta constraints, we create a cpmpy constraints `all(meta.constraints)` which is the exact translation of what the meta constraint is.


g-poveda (approved):
lgtm

PR #462, opened 2026-04-24
Add precedence mixins for tasks problem

nhuet:
- add a precedence mixin for solution to have automatic checks
- add precedence+scheduling mixins to implement the check on tasks order using start and end times
- create constraint generically in a dedicated cpsat mixin
- adapt implemented tasks problems

To avoid an ever growing and much too long discrete_optimization/generic_tasks_tools/solvers/cpsat.py, we split generic tasks solvers into subpackages cpsat and optal, mimicking problem/solution mixins structure


g-poveda (approved):
all good to me

PR #463, opened 2026-05-05
Create a generic "auto" cpsat solver for scheduling problems

nhuet:
- `GenericSchedulingAutoCpSatSolver` works for most generic problem class `GenericSchedulingProblem` with features
  - scheduling 
  - allocation
  - multimode
  - cumulative calendar resources 
  - non-renewable resources 
  - precedence constraints
- new mixins added to be able to adapt this solver to problem with less features without having to implement unnecessary method:
  - SinglemodeScheduling
  - WithoutNonRenewableResource
  - WithoutAllocation
  - WithoutCumulativeResource
  - WithoutRenewableResource
- always assume that this is at least a scheduling problem (start/end variables created)
- automatically create cpsat variables
- add constraints corresponding to the different features
- warmstart included
- generate temporary solution to be converted to actual problem solution via `convert_task_variables_to_solution()` to implement
- extra custom constraints to be added by overriding `init_model()`
- subclass `SinglemodeGenericSchedulingAutoCpSatSolver` gives access to `get_task_interval()` to avoid having to call a multimode method with mode=default_mode
- auto solver adapted for
  -  rcpsp
  - rcpsp_multiskill
  - jsp
  - fjsp
  - workforce/scheduling

### DIFF
--- a/src/discrete_optimization/fjsp/solvers/cpsat_auto.py
+++ b/src/discrete_optimization/fjsp/solvers/cpsat_auto.py
@@ -1,0 +1,116 @@
+#  Copyright (c) 2024 AIRBUS and its affiliates.
+#  This source code is licensed under the MIT license found in the
+#  LICENSE file in the root directory of this source tree.
+
+import logging
+from typing import Any
+
+from discrete_optimization.fjsp.problem import FJobShopProblem, FJobShopSolution, Task
+from discrete_optimization.generic_tasks_tools.allocation import (
+    NoUnaryResource,
+    UnaryResource,
+)
+from discrete_optimization.generic_tasks_tools.cumulative_resource import (
+    NoCumulativeResource,
+)
+from discrete_optimization.generic_tasks_tools.enums import StartOrEnd
+from discrete_optimization.generic_tasks_tools.non_renewable_resource import (
+    NoNonRenewableResource,
+)
+from discrete_optimization.generic_tasks_tools.solvers.cpsat.auto import (
+    GenericSchedulingAutoCpSatSolver,
+    TemporarySolution,
+)
+from discrete_optimization.generic_tools.hyperparameters.hyperparameter import (
+    CategoricalHyperparameter,
+)
+
+logger = logging.getLogger(__name__)
+
+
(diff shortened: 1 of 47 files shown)


src/discrete_optimization/workforce/scheduling/problem.py
```diff
@@ -105,10 +108,10 @@ def __init__(
 
 class AllocSchedulingProblem(
     AllocationSchedulingProblem[
-        Task, UnaryResource, CumulativeResource, NonRenewableResource
+        Task, UnaryResource, CumulativeResource, NoNonRenewableResource
     ],
-    WithoutNonRenewableResourceProblem[Task, NonRenewableResource],
-    SinglemodeProblem[Task],
+    WithoutNonRenewableResourceProblem[Task],
```
g-poveda, line 114: is it needed ? in the AllocationSchedulingProblem, by specifying "NoNonRenewableResource" isn't enough ? i guess you're right but just to be sure

src/discrete_optimization/jsp/problem.py
```diff
@@ -55,7 +68,11 @@ def __init__(self, machine_id: int, processing_time: int):
         self.processing_time = processing_time
 
 
-class JobShopProblem(PrecedenceSchedulingProblem[Task]):
+class JobShopProblem(
```
g-poveda, line 90: just an idea, i think jsp is somehow a renewable problem with unary resource (but we dont have the choice of the resource/machine) maybe a small rethinking would allow to have automatic implementation of the no_overlap ?

src/discrete_optimization/fjsp/solvers/cpsat_auto.py
```diff
@@ -0,0 +1,89 @@
...
+    TemporarySolution,
+)
+from discrete_optimization.generic_tools.hyperparameters.hyperparameter import (
+    CategoricalHyperparameter,
+)
+
+logger = logging.getLogger(__name__)
+
+
+class CpSatAutoFjspSolver(
```
g-poveda, line 31: same comment as jsp, maybe the cumulative could be also autocoded, but dont know how hard it would be

PR #466, opened 2026-05-06
Problem transformation functionality

g-poveda:
In this big PR, we add a new functionality around problem transformation.
-The idea is to add an api (in generic_tools/transformation) that is specifically done to transform a problem into another one, keeping either an exact or approximated mapping, and convert also the solution from 1 problem to another one.
-We then build a TransformationSolver class that is a new SolverDO class that take a source problem, a problem transformation and a solver subbrick to build a new solver. In practice it can be used for example to solve a binpack problem using a RCPSP solver.
-We introduce a new api idea to generalize SequentialMetaSolver, by creating a DAG of solver/transformation that can be used to solve a problem, it allows "parallel" solver tries (in different branches of the DAG) even though it's not done in parallel in practice.
-Added transformation for most problem and associated tests.


g-poveda:
closed and replace by #473

PR #473, opened 2026-05-13
Problem transformation functionality

g-poveda:
In this big PR, we add a new functionality around problem transformation.
-The idea is to add an api (in generic_tools/transformation) that is specifically done to transform a problem into another one, keeping either an exact or approximated mapping, and convert also the solution from 1 problem to another one.
-We then build a TransformationSolver class that is a new SolverDO class that take a source problem, a problem transformation and a solver subbrick to build a new solver. In practice it can be used for example to solve a binpack problem using a RCPSP solver.
-We introduce a new api idea to generalize SequentialMetaSolver, by creating a DAG of solver/transformation that can be used to solve a problem, it allows "parallel" solver tries (in different branches of the DAG) even though it's not done in parallel in practice.
-Added transformation for most problem and associated tests.
Replace #466

### DIFF
--- a/src/discrete_optimization/generic_tools/graph_solver/solver_graph.py
+++ b/src/discrete_optimization/generic_tools/graph_solver/solver_graph.py
@@ -6,10 +6,13 @@
 
 from __future__ import annotations
 
+import logging
 from abc import ABC, abstractmethod
-from collections import defaultdict, deque
+from collections import defaultdict
 from dataclasses import dataclass
 from typing import Any, Optional
+
+import networkx as nx
 
 from discrete_optimization.generic_tools.do_problem import (
     Problem,
@@ -24,6 +27,8 @@ from discrete_optimization.generic_tools.result_storage.result_storage import (
 from discrete_optimization.generic_tools.transformation.problem_transformation import (
     ProblemTransformation,
 )
+
+logger = logging.getLogger(__name__)
 
 
 @dataclass
@@ -410,6 +415,9 @@ class SolverGraph:
     # Execution state
     node_outputs: dict[str, NodeData]  # node_id -> NodeData
 
+    # NetworkX graph cache
+    _nx_graph: Optional[nx.DiGraph]  # Cached NetworkX representation
+
     def __init__(self, source_problem: Problem):
         """Initialize solver graph.
 
@@ -422,6 +430,7 @@ class SolverGraph:
         self.edges = defaultdict(list)
         self.reverse_edges = defaultdict(list)
         self.node_outputs = {}
+        self._nx_graph = None
 
     def add_transformation(
         self, node_id: str, transformation: ProblemTransformation
@@ -436,6 +445,9 @@ class SolverGraph:
             Node ID (for chaining)
 
         """
+        if node_id in self.nodes:
+            logging.error(f"{node_id} already added in the graph computation")
+            return None
         node = TransformationNode(node_id, transformation)
         self.nodes[node_id] = node
         return node_id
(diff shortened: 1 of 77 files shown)


src/discrete_optimization/generic_tools/graph_solver/solver_graph.py
```diff
@@ -0,0 +1,642 @@
...
+            node_id: Unique identifier for this node
+            transformation: Transformation to apply
+
+        Returns:
+            Node ID (for chaining)
+
+        """
+        node = TransformationNode(node_id, transformation)
+        self.nodes[node_id] = node
+        return node_id
```
nhuet, line 453: Slight risk: if user provides a node_id already used by mistake, a previous node could be forgotten. A possibility would be that the id is constructed by SolverGraph to ensure unicity and the user would only provide a meaningful name (that does not require to be unique).

src/discrete_optimization/generic_tools/graph_solver/solver_graph.py
```diff
@@ -0,0 +1,642 @@
...
+
+            for downstream in self.edges[node_id]:
+                in_degree[downstream] -= 1
+                if in_degree[downstream] == 0:
+                    queue.append(downstream)
+
+        if len(result) != len(self.nodes):
+            raise ValueError("Graph contains a cycle!")
+
+        return result
```
nhuet: Maybe use networkx to avoid recoding this sort of methods ?

src/discrete_optimization/generic_tools/transformation/problem_transformation.py
```diff
@@ -0,0 +1,216 @@
...
+
+        DEPRECATED: Use get_forward_metadata() and get_backward_metadata() instead.
+
+        Returns forward metadata by default for compatibility.
+
+        Returns:
+            TransformationMetadata (forward transformation)
+
+        """
+        return self.get_forward_metadata()
```
nhuet: How a completely new class in a completely new module could already have *deprecated* methods available for *compatibility* ?

src/discrete_optimization/generic_tools/transformation/solver_discovery.py
```diff
@@ -0,0 +1,399 @@
...
+
+def find_best_transformation_path(
+    source: str,
+    target: str,
+    transformation_graph: Optional[TransformationGraph] = None,
+) -> Optional[TransformationPath]:
+    """Find best transformation path between problems.
+
+    "Best" = prefer exact, then shortest, then lowest weight.
+
```
nhuet, line 375: It is actually "exact, then lowest weight, then shortest" according to code below

src/discrete_optimization/generic_tools/transformation/transformation_graph.py
```diff
@@ -0,0 +1,570 @@
...
+from enum import Enum
+from pathlib import Path
+from typing import Optional
+
+try:
+    import networkx as nx
+
+    NETWORKX_AVAILABLE = True
+except ImportError:
+    NETWORKX_AVAILABLE = False
```
nhuet: networkx is already in d-o dependencies, so this test should not be useful

src/discrete_optimization/generic_tools/graph_solver/solver_graph.py
```diff
@@ -0,0 +1,642 @@
...
+            )
+            # Use the problem's aggregation function (respects ParamsObjectiveFunction)
+            fit_value = self.aggreg_from_sol(sol_source)
+            output_result.append((sol_source, fit_value))
+
+        # Return NodeData (solution auto-extracted in __post_init__)
+        return NodeData(problem=self.source_problem, result=output_result)
+
+
+class SolverGraph:
```
nhuet, line 380: Never tested in unit tests. Only used in examples.

nhuet (commented):
Should be a very useful PR, unlocking a lot of possibilities and solvers at once.

PR #475, opened 2026-05-18
Add a generic CPM for scheduling problems to bounds task start/ends 

nhuet:
The "CPM" (inspired by
discrete_optimization.rcpsp.solvers.cpm.CpmRcpspSolver) propagate bounds
through precedence graph.
- forward: lower bounds, starting from 0 with starting tasks
- backward: upper bounds, starting from horizon with ending tasks

If we forget about other constraints and set horizon to
max(end_lower_bounds of ending tasks), we get classical cpm and are able
to extract a critical path.

This can be used by cpsat auto to precompute tighter start/ends bounds to restrict the space for start/end variables,
according to the setting `use_cpm_for_task_bounds`.

It was already done by fjsp cpsat solver so we set
`use_cpm_for_task_bounds=True` for the corresponding cpsat auto solver.

As it could be possibly long to compute for big rcpsp problems, we let
the setting to Flase by default for now.

We let also the possibility to use previously computed bounds with the
argument `task_bounds` to give directly in `init_model()`.

### DIFF
--- a/src/discrete_optimization/generic_tasks_tools/solvers/cpm.py
+++ b/src/discrete_optimization/generic_tasks_tools/solvers/cpm.py
@@ -1,7 +1,7 @@
-#  Copyright (c) 2022 AIRBUS and its affiliates.
+#  Copyright (c) 2026 AIRBUS and its affiliates.
 #  This source code is licensed under the MIT license found in the
 #  LICENSE file in the root directory of this source tree.
-"""CPM toc ompute lower/uppe bound on start/end of tasks"""
+"""CPM to compute lower/uppe bound on start/end of tasks"""
 
 import logging
 from dataclasses import dataclass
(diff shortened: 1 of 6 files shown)


src/discrete_optimization/generic_tasks_tools/solvers/cpm.py
```diff
@@ -0,0 +1,255 @@
...
+            for node in self.graph_nx.nodes()
+        }
+
+    def compute_task_bounds(self) -> None:
+        # Forward pass for lower bounds
+        # starting nodes: without ancestors, start lower bound = 0
+        queue = [
+            (0, self.node_to_index[node])
+            for node in self.graph_nx.nodes()
+            if len(nx.algorithms.ancestors(self.graph_nx, node)) == 0
```
g-poveda: could use self.predecessors ?

g-poveda (approved):

src/discrete_optimization/generic_tasks_tools/solvers/cpm.py
```diff
@@ -0,0 +1,255 @@
+#  Copyright (c) 2022 AIRBUS and its affiliates.
+#  This source code is licensed under the MIT license found in the
+#  LICENSE file in the root directory of this source tree.
+"""CPM toc ompute lower/uppe bound on start/end of tasks"""
```
g-poveda: to compute

src/discrete_optimization/generic_tasks_tools/solvers/cpm.py
```diff
@@ -0,0 +1,255 @@
+#  Copyright (c) 2022 AIRBUS and its affiliates.
```
g-poveda: 2026?

g-poveda (changes requested):

src/discrete_optimization/generic_tasks_tools/solvers/cpm.py
nhuet: Yes. Actually i wanted to remove self.predecessors from attributes as it was used only once. But i see i will have to use the reverse twice since i use also in critical_path. So i will definitely use self.predecessors

src/discrete_optimization/generic_tasks_tools/solvers/cpm.py
nhuet: The copy/paste from rcpsp/cpm... Normally pycharm fix it automatically but not here :)

PR #480, opened 2026-05-22
Improve cpsat auto (no optional interval + energy constraints)

nhuet:
Add option to avoid using optional intervals and add energy constraints

Inspired by https://github.com/google/or-tools/blob/6c55890/examples/python/rcpsp_sat.py we add options to:
- avoid using optional intervals (cpsat seems having more success solving scheduling problems)
- add energy constraints on nested subproblems (subset of tasks)

The options are enabled in `init_model()` (and added as hyperparameters) to ease d-o or optuna study creation.


g-poveda (approved):
all good to me as discussed

PR #494, opened 2026-06-15
Replace use of functools.cache/lru_cache by wrapt.lru_cache on methods

nhuet:
It avoids some pitfalls. See more about it here:
- https://wrapt.readthedocs.io/en/master/bundled.html#lru-cache
- https://docs.python.org/3/library/functools.html#functools.lru_cache
- https://docs.python.org/3/faq/programming.html#faq-cache-method-calls
- https://github.com/python/cpython/issues/102618

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -54,6 +54,7 @@ dependencies = [
     "didppy>=0.8.0",
     "setuptools",
     "pandas>=2",
+    "wrapt>=2.2.1",
 ]
 dynamic = ["version"]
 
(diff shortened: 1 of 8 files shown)


nhuet:
Some issues found in the way:
- https://github.com/GrahamDumpleton/wrapt/issues/342 -> solved in future version of wrapt 2.2.2rc1
- https://github.com/GrahamDumpleton/wrapt/issues/343

g-poveda (approved):
lgtm

PR #498, opened 2026-06-22
Improve generic scheduling problem/solver implementation

nhuet:
- fix `__getstate__`  so that it works with multiple inheritance
- change  "resource_consumptions" into "resources_levels"  as it is actually the min capacity needed that is optimized rather than the total consumption
- compute (optionally) a time penalty when evaluating generic scheduling solutions
- add a cost objective taking into account
  - a cost per mode (possibly related to the duration and resources consumption)
  - a cost per unary resource allocation (possibly related to the duration as the defined cost is per mode)


g-poveda (approved):
lgtm

PR #499, opened 2026-06-22
Shop floor problems refactoring

g-poveda:
-jobshop, flexiblejobshop and newly introduced openshop problems, now share a common base problem.
-cpsat is now only the auto implementation
-add a new scheduling mixin 'noOverlapProblem" to model the openshop problem -> each operation in a job can be executed in any order, but one at a time.
-update of example/tests


src/discrete_optimization/generic_tasks_tools/no_overlap_scheduling.py
```diff
@@ -0,0 +1,65 @@
...
+
+
+class NoOverlapProblem(SchedulingProblem[Task]):
+    """Problem with no overlap between tasks:
+    For example for open-shop problem, where the order of each operations
+    is arbitrary, but still no overlap.
+    """
+
+    @abstractmethod
+    def get_no_overlap(self) -> set[frozenset[Task]]:
```
nhuet, line 24: It is nice to specify this should be an immutable set. But there is a lot of other places in the library where we do not (using or sets or lists for problem characteristics).

src/discrete_optimization/flex_scheduling/problem.py
```diff
@@ -334,6 +334,9 @@ class FlexProblem(
     ],
     WithoutAllocationProblem[Task],
 ):
+    def get_no_overlap(self) -> set[frozenset[Task]]:
```
nhuet: Maybe use WithoutNoOverlap mixin ?

src/discrete_optimization/generic_tasks_tools/generic_scheduling_impl.py
```diff
@@ -128,6 +129,7 @@ def __init__(
             end_to_end_min_time_lags: min time lags constraints between task ends.
                 task1, task2, offset meaning end(task1) + offset <= end(task2)
                 Note that using negative offset can model end-to-end max time lags.
+            no_overlap_sets: a set of (set of tasks that should not overlap together)
```
nhuet, line 132: I thought at first it was a typo so could probably be clearer. Maybe "a set of sets of tasks that should not overlap together" is slightly better ?

src/discrete_optimization/rcpsp/problem.py
```diff
@@ -343,6 +347,10 @@ def update_resource_availabilities(self) -> None:
         super().update_resource_availabilities()
         self.get_resource_availabilities.cache_clear()
 
+    def get_no_overlap(self) -> set[frozenset[Task]]:
+        # In the future, it could make sense to put some tasks here (sharing same disjunctive constraint)
+        return {}
+
```
nhuet: Use rather WihtoutNoOverlap mixin

src/discrete_optimization/shop/fjsp/solvers/cpsat_auto.py
```diff
@@ -0,0 +1,12 @@
+#  Copyright (c) 2026 AIRBUS and its affiliates.
+#  This source code is licensed under the MIT license found in the
+#  LICENSE file in the root directory of this source tree.
+
+import logging
+
+from discrete_optimization.shop.solvers.cpsat_auto import CpSatAutoShopSolver
+
+logger = logging.getLogger(__name__)
```
nhuet: unused

src/discrete_optimization/shop/jsp/solvers/cpsat.py
```diff
@@ -0,0 +1,12 @@
+#  Copyright (c) 2026 AIRBUS and its affiliates.
+#  This source code is licensed under the MIT license found in the
+#  LICENSE file in the root directory of this source tree.
+
+import logging
+
+from discrete_optimization.shop.solvers.cpsat_auto import CpSatAutoShopSolver
+
+logger = logging.getLogger(__name__)
```
nhuet, line 9: not used

src/discrete_optimization/shop/jsp/solvers/cpsat_auto.py
```diff
@@ -0,0 +1,12 @@
...
+#  LICENSE file in the root directory of this source tree.
+
+import logging
+
+from discrete_optimization.shop.solvers.cpsat_auto import CpSatAutoShopSolver
+
+logger = logging.getLogger(__name__)
+
+
+CpSatAutoJspSolver = CpSatAutoShopSolver
```
nhuet: Now that we removed the old cpsat solver, i would rather call this one CpSatJspSolver for sake of simplicity. (to my mind the auto versions are meant to replace everywhere the old ones, litterally, i mean using the old name)

src/discrete_optimization/shop/jsp/problem.py
```diff
@@ -0,0 +1,50 @@
...
+    def get_precedence_constraints(self) -> dict[Task, list[Task]]:
+        return {
+            (j, k): [(j, k + 1)] if k + 1 < self.nb_subjob_per_job[j] else []
+            for j, job in enumerate(self.list_jobs)
+            for k in range(len(job.subjobs))
+        }
+
+    def get_no_overlap(self) -> set[frozenset[Task]]:
+        # Already taken into account in precedence.
+        return {}
```
nhuet: use WithoutNoOverlap ?

src/discrete_optimization/generic_tasks_tools/no_overlap_scheduling.py
```diff
@@ -0,0 +1,65 @@
...
+        """Map each task to the tasks that need to be performed after it."""
```
nhuet: wrong docstring

nhuet (commented):
Nice refactoring. Jsp and Fjsp were already sharing a lot indeed.
Now that you get rid of old cpsat solvers, i think it would be simpler to call the cpsat auto versions as the old ones (aka without including "auto" in the name). What do you think ?

src/discrete_optimization/generic_tasks_tools/no_overlap_scheduling.py
```diff
@@ -22,7 +22,10 @@ class NoOverlapProblem(SchedulingProblem[Task]):
 
     @abstractmethod
     def get_no_overlap(self) -> set[frozenset[Task]]:
-        """Map each task to the tasks that need to be performed after it."""
+        """
+        A set in this returned object is a (frozen) set of task,
+        where no task should overlap with another one in this set
+        """
```
nhuet, line 28: I think you meant "An object in this returned set is a (frozen) set of task, where no task should overlap with another one in this set" (so invert set <-> object)

nhuet (approved):

PR #505, opened 2026-07-08
Lot sizing implementation

g-poveda:
this PR introduce the lot sizing problem into the library
we build a generic api for specifying different variant of the problem:
-capacitated
-setup cost/times
-single/multi item
-with/without delays
-with/without changeorder
Following the concept of generic scheduling, a generic lot sizing implementation is done. 
Then some generic solver are also implemented (cpsat, milp and dp)
Some more specific implementation of problem and solvers are also provided for some variants of the problem, including the one specified in https://www.csplib.org/Problems/prob058/ where we replicate a simulated annealing custom solver, not using the generic sa framework of DO that has computational time issue


src/discrete_optimization/lotsizing/generic_solver/cpsat/lotsizing_solver_cpsat.py
nhuet, line 1: My small contribution: typo on this file name: lotsing -> lotsizing

src/discrete_optimization/lotsizing/generic_solver/milp/lotsizing_solver_milp.py
nhuet, line 1: Typo on the module name

PR #508, opened 2026-07-17
Make doctest failure only trigger warnings instead of workflow errors

nhuet:
- The doctest step always will be shown as successful.
- Pytest output is still kept to be looked upon in doctest step.
- Errors in doctest will be displayed as warnings in workflow summary.
- A job summary is generated to have proper links to code and log (as
  default warnings links from github are not very meaningful)

Notes:
- step name is not available from within it => we need to pass it as an
  env variable
- job name as generated per the matrix is not available from within the
  step, only th job id => we generate the name ourself and duplicate the
  generation in a env variable
- we need the step number inside the job (number created during the job,
  taking into account preparation steps added by github) => we need to
  make agithub api request to find it, given the job name and the step
  name

### DIFF
--- a/.github/scripts/parse_doctest.py
+++ b/.github/scripts/parse_doctest.py
@@ -1,0 +1,109 @@
+import json
+import os
+import re
+import sys
+import urllib.request
+
+
+def get_log_url(log_line_number):
+    """Fetches exact runtime Job ID and Step Number dynamically using GitHub APIs"""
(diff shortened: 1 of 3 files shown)


nhuet:
See example of generated summary in case of docstrings errors here: 
- push on a branch: https://github.com/nhuet/discrete-optimization/actions/runs/29832683894
- push on a PR: https://github.com/nhuet/discrete-optimization/actions/runs/29832687940

PR #518, opened 2026-08-18
Generic scheduling API additionnal features

g-poveda:
- we add richer input to the genericschedulingproblem in order to model more existing scheduling problem in the repository:
- Same Unary Allocation constraint : allow to specify set of task that should be allocated to the same unary resource
- ModeConstraint : allow to specify implication/cascade of constraints between modes of subset of tasks.
- Dispersion of workload objective generic implementation
- TaskEntity creation : 
     -  we can create entities corresponding to group of tasks. 
     -  this is used to create some constraint called resource blocking : some resource can be considered as used between start/end of 2 entities. or during the span of a given entity.


src/discrete_optimization/binpack/transformations/to_generic.py
```diff
@@ -0,0 +1,148 @@
...
+    exact_transformation,
+)
+
+
+class BinpackToGenericApproach(Enum):
+    ALLOCATION = 0
+    SCHEDULING = 1
+
+
+class BinpackToGenericSchedulingTransformation(
```
nhuet, line 38: You do not want to take advantage of the generic abstract class ToGenericSchedulingImpl ? 
In particular it gives you for free the `transform_problem()` (and allow to maitain it in one single place). May be you still will have to override it to add the check `self.modeling == BinpackToGenericApproach.SCHEDULING` but then only call `super().transform_problem()`. I guess that this check should also be added to the implemtation of `transform_solution_from_raw_generic_to_specific()` used for `back_transform_solution()`.
However i am not sure why this attribute `modeling` in the transformer? As the transformer name suggests that its goal is already to transform to *scheduling* generic pb and not *allocation* generic pb.

g-poveda: This Binpack problem can be seen as a scheduling problem with a few constraint (like no_overlap for items being in conflict). Though it's not really the natural representation for this problem, so i didn't make BinpackProblem as a subclass of GenericSchedulingProblem+Mixins, as it would force to implement some function/property that is not so natural for binpacking. thus this "manual" transformation, what do u think

PR #520, opened 2026-08-19
Scheduling problem with Resource consumption dependent on other task modes

g-poveda:
we introduce a new feature to the problem, the possibility that the resource consumption of task in a given mode, might depend not only on the task mode, but also to the mode values of other task. This can model problem where decision done in early stage has an impact on future task. 

For example, 
-if we choose to accomplish task0 in a given location, it might impact the consumption of further task1 if it is implicitely done in the same place.
-Or if we use some mode of task0, then the "cost" to accomplish a further task can be increased/decreased. 
We add new checker, cpsat model and a problem implementing this feature.

### DIFF
--- a/src/discrete_optimization/generic_tasks_tools/__init__.py
+++ b/src/discrete_optimization/generic_tasks_tools/__init__.py
@@ -1,0 +1,8 @@
+#  Copyright (c) 2026 AIRBUS and its affiliates.
+#  This source code is licensed under the MIT license found in the
+#  LICENSE file in the root directory of this source tree.
+from enum import Enum
+
+
+class UndefinedInput(Enum):
+    UNDEFINED = 0
--- a/src/discrete_optimization/generic_tasks_tools/__init__.py
+++ b/src/discrete_optimization/generic_tasks_tools/__init__.py
@@ -1,8 +1,3 @@
 #  Copyright (c) 2026 AIRBUS and its affiliates.
 #  This source code is licensed under the MIT license found in the
 #  LICENSE file in the root directory of this source tree.
-from enum import Enum
-
-
-class UndefinedInput(Enum):
-    UNDEFINED = 0
--- a/src/discrete_optimization/generic_tasks_tools/cumulative_resource.py
+++ b/src/discrete_optimization/generic_tasks_tools/cumulative_resource.py
@@ -40,12 +40,13 @@ class CumulativeResourceProblem(
         self, resource: CumulativeResource, task: Task, mode: int
     ) -> dict[frozenset[tuple[Task, int]], int]:
         # To be Overridden in child classes
+
         return {}
 
-    def get_possible_resource_consumption(
+    def get_possible_cumulative_resource_consumption(
         self, resource: CumulativeResource, task: Task, mode: int
     ) -> set[int]:
-        if self.is_resource_task_mode_consumption_dependent(
+        if self.is_cumulative_resource_task_mode_consumption_dependent(
             resource=resource, task=task, mode=mode
         ):
             return set(
@@ -59,12 +60,12 @@ class CumulativeResourceProblem(
             )
         }
 
-    def get_possible_resource_consumption_all_modes(
+    def get_possible_cumulative_resource_consumption_all_modes(
         self, resource: CumulativeResource, task: Task
     ) -> set[int]:
         return reduce(
             lambda prev, y: prev.union(
-                self.get_possible_resource_consumption(
+                self.get_possible_cumulative_resource_consumption(
                     resource=resource, task=task, mode=y
                 )
             ),
@@ -72,31 +73,36 @@ class CumulativeResourceProblem(
             set(),
         )
 
-    def is_resource_task_mode_consumption_dependent(
+    def is_cumulative_resource_task_mode_consumption_dependent(
         self, resource: CumulativeResource, task: Task, mode: int
     ) -> bool:
         # To be Overridden in child classes
         return False
 
-    def is_resource_task_consumption_dependent(
+    def is_cumulative_resource_task_consumption_dependent(
         self, resource: CumulativeResource, task: Task
     ):
         return any(
-            self.is_resource_task_mode_consumption_dependent(
+            self.is_cumulative_resource_task_mode_consumption_dependent(
                 resource=resource, task=task, mode=mode
             )
             for mode in self.get_task_modes(task)
         )
 
-    def is_task_consumption_dependent(self, task: Task):
+    def is_task_cumulative_consumption_dependent(self, task: Task):
         # To be Overridden in child classes
         return any(
-            self.is_resource_task_consumption_dependent(resource=resource, task=task)
+            self.is_cumulative_resource_task_consumption_dependent(
+                resource=resource, task=task
+            )
             for resource in self.cumulative_resources_list
         )
 
-    def has_any_consumption_dependent(self):
-        return any(self.is_task_consumption_dependent(task) for task in self.tasks_list)
+    def has_any_cumulative_consumption_dependent(self):
+        return any(
+            self.is_task_cumulative_consumption_dependent(task)
+            for task in self.tasks_list
+        )
 
     @abstractmethod
     def get_cumulative_resource_consumption(
@@ -175,7 +181,7 @@ class CumulativeResourceSolution(
 
         """
         if self.problem.is_cumulative_resource(resource):
-            if not self.problem.is_resource_task_mode_consumption_dependent(
+            if not self.problem.is_cumulative_resource_task_mode_consumption_dependent(
                 resource, task, self.get_mode(task)
             ):
                 return self.problem.get_cumulative_resource_consumption(
--- a/src/discrete_optimization/generic_tasks_tools/cumulative_resource.py
+++ b/src/discrete_optimization/generic_tasks_tools/cumulative_resource.py
@@ -3,7 +3,9 @@
 #  LICENSE file in the root directory of this source tree.
 from __future__ import annotations
 
+import logging
 from abc import abstractmethod
+from functools import reduce
 from typing import Generic, Hashable, TypeVar
 
 from discrete_optimization.generic_tasks_tools.base import Task
@@ -33,6 +35,68 @@ class CumulativeResourceProblem(
     The task consumption of these cumulative resources is supposed to be determined entirely determined by the task mode.
 
     """
+
+    def get_cumulative_resource_consumption_mapping(
+        self, resource: CumulativeResource, task: Task, mode: int
+    ) -> dict[frozenset[tuple[Task, int]], int]:
+        # To be Overridden in child classes
+        return {}
+
+    def get_possible_resource_consumption(
+        self, resource: CumulativeResource, task: Task, mode: int
+    ) -> set[int]:
+        if self.is_resource_task_mode_consumption_dependent(
+            resource=resource, task=task, mode=mode
+        ):
+            return set(
+                self.get_cumulative_resource_consumption_mapping(
+                    resource=resource, task=task, mode=mode
+                ).values()
+            )
+        return {
+            self.get_cumulative_resource_consumption(
+                resource=resource, task=task, mode=mode
+            )
+        }
+
+    def get_possible_resource_consumption_all_modes(
+        self, resource: CumulativeResource, task: Task
+    ) -> set[int]:
+        return reduce(
+            lambda prev, y: prev.union(
+                self.get_possible_resource_consumption(
+                    resource=resource, task=task, mode=y
+                )
+            ),
+            list(self.get_task_modes(task)),
+            set(),
+        )
+
+    def is_resource_task_mode_consumption_dependent(
+        self, resource: CumulativeResource, task: Task, mode: int
+    ) -> bool:
+        # To be Overridden in child classes
+        return False
+
+    def is_resource_task_consumption_dependent(
+        self, resource: CumulativeResource, task: Task
+    ):
+        return any(
+            self.is_resource_task_mode_consumption_dependent(
+                resource=resource, task=task, mode=mode
+            )
+            for mode in self.get_task_modes(task)
+        )
+
+    def is_task_consumption_dependent(self, task: Task):
+        # To be Overridden in child classes
+        return any(
+            self.is_resource_task_consumption_dependent(resource=resource, task=task)
+            for resource in self.cumulative_resources_list
+        )
+
+    def has_any_consumption_dependent(self):
+        return any(self.is_task_consumption_dependent(task) for task in self.tasks_list)
 
     @abstractmethod
     def get_cumulative_resource_consumption(
@@ -76,6 +140,28 @@ class CumulativeResourceSolution(
 
     problem: CumulativeResourceProblem[Task, CumulativeResource, OtherCalendarResource]
 
+    def get_calendar_resource_consumption_from_mapping(
+        self, resource: CumulativeResource, task: Task
+    ) -> int:
+        mode = self.get_mode(task)
+        mapping = self.problem.get_cumulative_resource_consumption_mapping(
+            resource=resource, task=task, mode=mode
+        )
+        set_of_tasks = set([frozenset([t for t, m in k]) for k in mapping])
+        value = next(
+            (
+                mapping[key_mapping]
+                for set_task in set_of_tasks
+                if (key_mapping := frozenset([(t, self.get_mode(t)) for t in set_task]))
+                in mapping
+            ),
+            None,
+        )
+        if value is None:
+            logging.info(f"No found mapping")
+            return None
+        return value
+
     def get_calendar_resource_consumption(self, resource: Resource, task: Task) -> int:
         """Get resource consumption by given task.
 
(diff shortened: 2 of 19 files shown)


src/discrete_optimization/generic_tasks_tools/cumulative_resource.py
```diff
@@ -26,14 +28,133 @@ class CumulativeResourceProblem(
...
+    **Resource-Dependent Example:**
+
+    Task A's electricity varies based on whether task B's heater is on:
+
+        >>> mapping = {
+        ...     frozenset([(task_B, 0)]): 10,  # B heater off: A uses 10 kW
+        ...     frozenset([(task_B, 1)]): 8,   # B heater on: A uses 8 kW
+        ... }
+
+    Implementation pattern:
```
nhuet: Example:

src/discrete_optimization/generic_tasks_tools/cumulative_resource.py
```diff
@@ -76,22 +197,70 @@ class CumulativeResourceSolution(
...
+                mapping[key_mapping]
+                for set_task in set_of_tasks
+                if (key_mapping := frozenset([(t, self.get_mode(t)) for t in set_task]))
+                in mapping
+            ),
+            None,
+        )
+        if value is None:
+            logging.info(f"No found mapping")
+            return None
```
nhuet: from signature, we cannot return None. So we should either raise an error, or return 0.

src/discrete_optimization/generic_tasks_tools/generic_scheduling_impl.py
```diff
@@ -234,6 +243,48 @@ def check_resources_lists(self) -> None:
...
+    ) -> dict[frozenset[tuple[Task, int]], int]:
+        if self.is_cumulative_resource_task_mode_consumption_dependent(
+            resource, task, mode
+        ):
+            return self.resource_consumptions_dependent[task][mode][resource]
+        return {
+            frozenset([]): self.get_cumulative_resource_consumption(
+                resource, task, mode
+            )
+        }
```
nhuet: Why not return `super().get_cumulative_resource_consumption_mapping()` in that case? If the default implementation is not ok, maybe this implementation should be the default one.

src/discrete_optimization/generic_tasks_tools/cumulative_resource.py
```diff
@@ -26,14 +28,133 @@ class CumulativeResourceProblem(
...
+        ...             return self.mode_details[task][mode][resource]
+        ...         return {frozenset([]): self.get_cumulative_resource_consumption(
+        ...             resource, task, mode
+        ...         )}
+
+    See `src/discrete_optimization/rcpsp_resource_dependent/problem.py` for a complete implementation.
 
     """
 
+    def get_cumulative_resource_consumption_mapping(
```
nhuet, line 77: add `@optional_override` decorator

src/discrete_optimization/generic_tasks_tools/cumulative_resource.py
```diff
@@ -26,14 +28,133 @@ class CumulativeResourceProblem(
...
+            lambda prev, y: prev.union(
+                self.get_possible_cumulative_resource_consumption(
+                    resource=resource, task=task, mode=y
+                )
+            ),
+            list(self.get_task_modes(task)),
+            set(),
+        )
+
+    def is_cumulative_resource_task_mode_consumption_dependent(
```
nhuet, line 137: add `@optional_override` decorator

src/discrete_optimization/generic_tasks_tools/non_renewable_resource.py
```diff
@@ -68,12 +69,102 @@ def get_non_renewable_resource_consumption(
         """
         ...
 
+    def get_non_renewable_resource_consumption_mapping(
```
nhuet, line 80: add `@optional_override` decorator

src/discrete_optimization/generic_tasks_tools/non_renewable_resource.py
```diff
@@ -68,12 +69,102 @@ def get_non_renewable_resource_consumption(
...
+            lambda prev, y: prev.union(
+                self.get_possible_non_renewable_resource_consumption(
+                    resource=resource, task=task, mode=y
+                )
+            ),
+            list(self.get_task_modes(task)),
+            set(),
+        )
+
+    def is_non_renewable_resource_task_mode_consumption_dependent(
```
nhuet, line 120: add `@optional_override` decorator

src/discrete_optimization/generic_tasks_tools/non_renewable_resource.py
```diff
@@ -68,12 +69,102 @@ def get_non_renewable_resource_consumption(
...
+    ):
+        return any(
+            self.is_non_renewable_resource_task_mode_consumption_dependent(
+                resource=resource, task=task, mode=mode
+            )
+            for mode in self.get_task_modes(task)
+        )
+
+    def is_task_non_renewable_consumption_dependent(self, task: Task):
+        # To be Overridden in child classes
```
nhuet: Are you sure of this comment? It is probably `is_non_renewable_resource_task_mode_consumption_dependent` that is more likely to be overriden even though this is also possible.

src/discrete_optimization/generic_tasks_tools/non_renewable_resource.py
```diff
@@ -68,12 +69,102 @@ def get_non_renewable_resource_consumption(
...
+                for set_task in set_of_tasks
+                if (key_mapping := frozenset([(t, self.get_mode(t)) for t in set_task]))
+                in mapping
+            ),
+            None,
+        )
+        if value is None:
+            logging.info(f"No found mapping")
+            return None
+        return value
```
nhuet, line 175: from signature, we cannot return None. So we should either raise an error, or return 0.

src/discrete_optimization/generic_tasks_tools/generic_scheduling_impl.py
```diff
@@ -234,6 +243,48 @@ def check_resources_lists(self) -> None:
...
+    ) -> dict[frozenset[tuple[Task, int]], int]:
+        if self.is_non_renewable_resource_task_mode_consumption_dependent(
+            resource, task, mode
+        ):
+            return self.resource_consumptions_dependent[task][mode][resource]
+        return {
+            frozenset([]): self.get_non_renewable_resource_consumption(
+                resource, task, mode
+            )
+        }
```
nhuet: Why not return `super().get_non_renewable_resource_consumption_mapping()` in that case? If the default implementation is not ok, maybe this implementation should be the default one.

src/discrete_optimization/generic_tasks_tools/generic_scheduling_impl.py
```diff
@@ -234,6 +243,48 @@ def check_resources_lists(self) -> None:
...
+        except KeyError:
+            return False
+
+    def get_non_renewable_resource_consumption_mapping(
+        self, resource: CumulativeResource, task: Task, mode: int
+    ) -> dict[frozenset[tuple[Task, int]], int]:
+        if self.is_non_renewable_resource_task_mode_consumption_dependent(
+            resource, task, mode
+        ):
+            return self.resource_consumptions_dependent[task][mode][resource]
```
nhuet, line 277: maybe we can add a fallback to `{frozenset(): 0}` when key is not present ? as for `get_non_renewable_resource_consumption()` implementation. (And if as suggested below this is the default implementation in the mixin we could call `super().get_non_renewable_resource_consumption_mapping()`.

src/discrete_optimization/generic_tasks_tools/generic_scheduling_impl.py
```diff
@@ -234,6 +243,48 @@ def check_resources_lists(self) -> None:
...
+        except KeyError:
+            return False
+
+    def get_cumulative_resource_consumption_mapping(
+        self, resource: CumulativeResource, task: Task, mode: int
+    ) -> dict[frozenset[tuple[Task, int]], int]:
+        if self.is_cumulative_resource_task_mode_consumption_dependent(
+            resource, task, mode
+        ):
+            return self.resource_consumptions_dependent[task][mode][resource]
```
nhuet, line 260: maybe we can add a fallback to `{frozenset(): 0}` when key is not present ? as for `get_cumulative_resource_consumption()` implementation. (And if as suggested below this is the default implementation in the mixin we could call `super().get_cumulative_resource_consumption_mapping()`.

src/discrete_optimization/generic_tasks_tools/__init__.py
```diff
@@ -0,0 +1,8 @@
+#  Copyright (c) 2026 AIRBUS and its affiliates.
+#  This source code is licensed under the MIT license found in the
+#  LICENSE file in the root directory of this source tree.
+from enum import Enum
+
+
+class UndefinedInput(Enum):
+    UNDEFINED = 0
```
nhuet: Probably something from the "optional task" PR ?

src/discrete_optimization/generic_tasks_tools/transformations/generic_scheduling_impl.py
```diff
@@ -358,6 +343,84 @@ def _construct_min_only_time_lags_from_min_and_max_time_lags(
...
+            # Non-renewable resources
+            for resource in source_problem.non_renewable_resources_list:
+                if source_problem.is_non_renewable_resource_task_mode_consumption_dependent(
+                    resource=resource, task=task, mode=mode
+                ):
+                    resources_consumption_dependent[task][mode][resource] = (
+                        source_problem.get_non_renewable_resource_consumption_mapping(
+                            resource=resource, task=task, mode=mode
+                        )
+                    )
```
nhuet, line 396: as below we could remove keys for which the conso is 0. Under the condition that `GenericSchedulingImplProblem.get_non_renewable_resource_consumption_mapping()` include default to 0 when key is missing. (see comment in `GenericSchedulingImplProblem` code)

src/discrete_optimization/generic_tasks_tools/transformations/generic_scheduling_impl.py
```diff
@@ -358,6 +343,84 @@ def _construct_min_only_time_lags_from_min_and_max_time_lags(
...
+            # Cumulative resources
+            for resource in source_problem.cumulative_resources_list:
+                if source_problem.is_cumulative_resource_task_mode_consumption_dependent(
+                    resource=resource, task=task, mode=mode
+                ):
+                    resources_consumption_dependent[task][mode][resource] = (
+                        source_problem.get_cumulative_resource_consumption_mapping(
+                            resource=resource, task=task, mode=mode
+                        )
+                    )
```
nhuet, line 413: idem, may remove keys for conso=0 if consistentwith `GenericSchedulingImplProblem` code.

src/discrete_optimization/rcpsp_resource_dependent/solvers/cpsat.py
```diff
@@ -0,0 +1,62 @@
...
+
+class CpSatRcpspResourceDependentSolver(
+    GenericSchedulingAutoCpSatSolver[
+        Task, NoUnaryResource, NoSkill, NonSkillCumulativeResource, NonRenewableResource
+    ]
+):
+    problem: RcpspResourceDependentProblem
+
+    def needs_duration_variables(self) -> bool:
+        return True
```
nhuet: If this is needed while `init_model` is not overriden , this means that the default implementation in `needs_duration_variables` is faulty in `GenericSchedulingAutoCpSatSolver`, and should be changed in this PR.

src/discrete_optimization/generic_tasks_tools/solvers/cpsat/non_renewable_resource.py
```diff
@@ -30,7 +125,9 @@ def get_non_renewable_resource_demand_variable(
 
         Default to a linear expression using consumption per mode and is_present variables.
         If demand variables are indeed created in the cp_model, this should be overriden to return it
-        so that non-renewable resource constraints are constraining these variables.
+        so that cumulative resource constraints are constraining these variables.
+
+        Needed if `self.avoid_interval_optional` is set to True.
```
nhuet: wrong change of docstring

src/discrete_optimization/generic_tasks_tools/solvers/cpsat/non_renewable_resource.py
```diff
@@ -22,6 +23,100 @@ class NonRenewableCpSatSolver(
...
+                                resource=resource, task=task, mode=mode
+                            )
+                            self.cp_model.add(
+                                self.demands_non_renewable_resource_vars[task, resource]
+                                == value
+                            ).only_enforce_if(
+                                self.get_task_mode_is_present_variable(
+                                    task=task, mode=mode
+                                )
+                            )
```
nhuet: As `use_enforce_if_instead_of_sum` is True by default, this is a change of behaviour for demand variables even without dependent resource consumption. (So perhaps needs a little justification)µ

**But**, it is worth noticing that this method is overriden in`GenericSchedulingAutoCpSatSolver` where an other option (accessible in `init_model`) `use_demand_variables_for_non_renewable_resources` is set to False by default. If it is set to `True` however, we create *other* non-renewable resources demand variables, which do not take into account the dependency in resource consumption...

The two codes should be merged somehow to avoid that. Maybe by putting `use_demand_variables_for_non_renewable_resources` at the level of this mixin ? and removing the corresponding demand variables from `GenericSchedulingAutoCpSatSolver`?

But you should also take care to reset the initialization when `GenericSchedulingAutoCpSatSolver.init_model()` is called (by modifiying the method `_reset_variables()` which could call an equivalent in the parent mixins).

src/discrete_optimization/generic_tasks_tools/solvers/cpsat/auto.py
```diff
@@ -182,6 +182,8 @@ class GenericSchedulingAutoCpSatSolver(
     """Variables tracking level (capacity needed) of each (unary, cumulative, or non-renewable) resource."""
     resource_level_variables_created = False
     """Flag telling whether 'resource_level_variables' have been created"""
+    use_enforce_if_instead_of_sum = True
+    """For demand variable, use enforce if instead of sum of mode boolean*conso"""
```
nhuet: Technically this is not in the scope of this PR :) as it changes the behaviour of demand variables when no dependency exist for resource consumption.
And as this is a change to previous behaviour so maybe a little explanation about why this is more efficient to "enforce if instead of sum" could be nice.

src/discrete_optimization/generic_tasks_tools/solvers/cpsat/cumulative_resource.py
```diff
@@ -72,6 +78,97 @@ def get_resource_consumption_intervals(
...
+                                resource=resource, task=task, mode=mode
+                            )
+                            self.cp_model.add(
+                                self.demands_cumulative_resource_vars[task, resource]
+                                == value
+                            ).only_enforce_if(
+                                self.get_task_mode_is_present_variable(
+                                    task=task, mode=mode
+                                )
+                            )
```
nhuet: could reuse the utility method `GenericSchedulingAutoCpSatSolver._create_var_per_mode()`  (which will need to be put in a utils module as a function)

src/discrete_optimization/generic_tasks_tools/solvers/cpsat/non_renewable_resource.py
nhuet: could reuse the utility method GenericSchedulingAutoCpSatSolver._create_var_per_mode() (which will need to be put in a utils module as a function)

src/discrete_optimization/generic_tasks_tools/solvers/cpsat/auto.py
```diff
@@ -1117,8 +1121,11 @@ def _create_on_the_fly_task_interval(self, task: Task) -> None:
...
         self, task: Task, resource: CumulativeResource
     ) -> LinearExprT:
+        # TODO : FIX THIS BEHAVIOUR WHEN RESOURCE DEPENDENT.
         if self.avoid_interval_optional_for_cumulative_resources:
             return self.demand_variables[task][resource]
+        # if self.avoid_interval_optional:
+        #    return self.demand_variables[task][resource]
         return super().get_cumulative_resource_demand_variable(
             task=task, resource=resource
         )
```
nhuet: Maybe stop overriding it and rely only on `CumulativeResourceSchedulingCpSatSolver.get_cumulative_resource_demand_variable()` ? (which already do the job)
And remove the dict `self.demand_variables` not needed anymore.

(and then remove these 3 commented lines :)

Similarly call do not override anymore `get_non_renewable_resource_demand_variable` and remove not needed anymore `use_demand_variables_for_non_renewable_resources` attribute. 
And add `use_enforce_if_instead_of_sum` to hyperparameters instead of it.

nhuet (changes requested):

src/discrete_optimization/rcpsp_resource_dependent/solvers/cpsat.py
nhuet: Actually, looking the code, it seems that task intervals are actually needed if dependent resources are present.
So it is `needs_task_intervals()` that should be modiied to include `or self.problem.has_any_cumulative_consumption_dependent()`

nhuet (changes requested):

PR #523, opened 2026-08-24
Refactor objective handling for scheduling problem

g-poveda:


src/discrete_optimization/generic_tasks_tools/solvers/cpsat/auto.py
```diff
@@ -206,6 +202,9 @@ class GenericSchedulingAutoCpSatSolver(
     demand_non_renewable_modeling: ModeToValueModeling = ModeToValueModeling.ENFORCE_IF
     """Specify how the demand vars are defined with the modes"""
 
+    list_obj_modeler_weight: list[tuple[ObjectiveModelerCpSat, float]] = None
```
nhuet: Should not be initialized here i think

src/discrete_optimization/generic_tasks_tools/solvers/cpsat/auto.py
```diff
@@ -419,6 +418,7 @@ def _reset_variables(self):
         self.all_used_variables = {}
         self.resource_level_variables_created = False
         self.resource_level_variables = {}
+        self.list_obj_modeler_weight = None
```
nhuet: rather ` = [] ` according to the typing ? Makes more sense also.

src/discrete_optimization/generic_tasks_tools/solvers/cpsat/objectives/utils.py
```diff
@@ -0,0 +1,100 @@
+#  Copyright (c) 2026 AIRBUS and its affiliates.
+#  This source code is licensed under the MIT license found in the
+#  LICENSE file in the root directory of this source tree.
+
+
+def create_computer_to_modeler_mapping():
```
nhuet: Why a function rather than a global dict variable ? The + could be that later we allow customization of the mapping by the user...

nhuet:
Except for my minor comments, nice PR allowing plenty of new objectives off the shelf!

src/discrete_optimization/generic_tasks_tools/solvers/cpsat/auto.py
g-poveda: done in next commit, thx

PR #524, opened 2026-08-28
Add non-regression tests on cpsat solvers for some scheduling pbs

nhuet:
Classes of problems concerned by the non-regression tests:
- rcpsp
- rcpsp_ms
- jsp
- fjsp
- workforce/scheduling

Some utility fixtures are defined in tests/conftest.py to ease the creation of new non-regression tests.

Per test-module it defines
- a csv database file (module test name + _db.csv"
- a check_nonregression_fn function, which either tests against previous test runs found in the csv (default behaviour) or populate the csv with the current version of problems/solvers. You need for that to define and set  `NONREGRESSION_POPULATE_DATABASE=True` in the concerned test module.
- this checks that
  - optimal cases stay optimal with same objective values
  - satified cases are either improved or not worse that the stored objectives + margin (5% by default). You can set the margin to a different value by setting a variable `NONREGRESSION_MARGIN_OBJECTIVE` at test module level.
  - in case of multiobjectives, it is checked that the first components are equal and the next one is below the stored value + margin
- the csv database also stores the optimization sense (either maximization or minimzation) to translate properly the previous checks in case of maximization
- the usecase are identified by the id given by pytest or pytest-cases and retrieve via `request.node.nodeid`

As for doctests, we also do not fail the github actions in case of regression but rather raise warnings and export a csv compiling regressions in artifacts.
We also parse errors during collections of tests and add this improvement to the parsing of doctests.

See 
- an example with regressions and failures in some tests: https://github.com/nhuet/discrete-optimization/actions/runs/33494150762
- an example with collection errors: https://github.com/nhuet/discrete-optimization/actions/runs/33494132948


g-poveda (approved):
great, thx

PR #534, opened 2026-09-04
Update regression case for workforce scheduling

g-poveda:
- we identified instance_170 as one "unlucky" instance where the cpsat auto was regressing. Though, through deeper and larger studies (https://github.com/g-poveda/discrete-optimization/blob/gpd/investigate_reg/examples/workforce/scheduling/investigate_reg.py) we are confident that it's actually not a regression of the cpsat auto.


nhuet:
There are still regressions but it is improving :)
- ubuntu: no more regressions!
- macos: regressions on instance 147: 6, 150 instead of 4, 12 ...
- windows: not poping since #532 not yet merged, but 
  - instance 147: 5,67 au lieu de 4, 12
  - instance 252: 10,26 au lieu de 9, 84

nhuet (approved):
Nice it seems to work.

PR #540, opened 2026-09-15
Introduce optional tasks in tasks problems

nhuet:
- `TaskProblem` can define optional tasks (default to none)
- `TaskSolution.is_present()` tells if a task is present or absent in the solution
- in a solution, task mode,start and end can have a special value `AbsentValue.ABSENT`. It should be possible only if `sol.is_present(task)` is false.
- default `is_present()` implementations:
  - allocation: at least one resource allocated  (to be improved later according to pb parameter, for instance via at least/exactly/at most one resource flags per task)
  - mode: a mode is chosen (int value)
  - scheduling: start and end are defined (int values)
  - generic-scheduling: start, end and mode are defined (int values)
  - redefined in subproblems
- constraints and "objective" variables conditioned by presence/absence of tasks:
  - makespan: take only present tasks into account
  - nb resources used: idem
  - timelags/precedence constraints are ignored if one of the task is optional and absent
  - => cpsat variables is_present_mode, is_present_unary_resource, task_interval are conditioned by a boolean var telling if a given (optional) task is present. Start and end variables value are set to 0 when the task is absent.
- resource_blocking & entities: entities are active when one of their task is active (+ proper mode chosen for `TaskModeEntity`)
- new pseudo abstractmethod in `TasksCpSatSolver`: `get_task_is_present_variable()`, default to returning 1 for mandatory tasks, to be overriden for optional tasks (implemented in cpsat-auto)
- to avoid confusion, objective "nb_tasks_done" is renamed "nb_tasks_allocated"


g-poveda:
all good to me, fix some bugs with optional task + resource blocking feature: should be fixed in : 
https://github.com/g-poveda/discrete-optimization/commit/252ed5ebc007ff4f065abe02bf2a8f18cd65c84e i'll do. a PR just after

